PR #63, opened 2015-05-10
Add urls to events, fixes according to the specs, PEP8 fixes

jammon:
This pull request adds a thing and fixes many little details.

- Added support for URLs to Event.
- Fixes according to RFC5545:
    + [Lines are separated by CRLF (`\r\n`)](https://tools.ietf.org/html/rfc5545#section-3.1). Tests had to be adapted,  `str.split('\n')` is changed to the more generic `str.splitlines()`
    + [names of properties etc. are case-insensitive](https://tools.ietf.org/html/rfc5545#page-11). So in `ContentLine.__init__` `name` is uppercased for consistency and easier comparing. Some tests had to be adapted.
- PEP8 Fixes:
    + assigning lambdas to a variable is discouraged, (In eventlist.py there are a lot of these assignments left. I didn't work through this code yet.)
    + formating issues.
- Other fixes:
    + Use `str.split(splitChar, 1)` where appropriate
    + use `isinstance(foo, string_types)` instead of `isinstance(foo, str)`

### DIFF
--- a/ics/event.py
+++ b/ics/event.py
@@ -281,13 +281,12 @@ class Event(Component):
         """
         Returns:
             int: hash of self. Based on self.uid."""
-        ord3 = lambda x: '%.3d' % ord(x)
-        return int(''.join(map(ord3, self.uid)))
+        return int(''.join(map(lambda x: '%.3d' % ord(x), self.uid)))
 
 
-######################
-####### Inputs #######
-
+# ------------------
+# ----- Inputs -----
+# ------------------
 @Event._extracts('DTSTAMP')
 def created(event, line):
     if line:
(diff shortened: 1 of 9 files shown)


C4ptainCrunch:
Wow, thanks for this pull requests, it looks awesome :)
(I'll dig further into it today, expect a merge soon)

jammon:
Great, thank you!
(My first real contribution to an open source project :-)

C4ptainCrunch:
Do you need those changes to be pip-installable immediately ? I can do a pypi release if you want :)

jammon:
:)
Thanks, I did all these changes for my own purpose, so for me it's ok to have them on my machine. Do a pypi release whenever you deem it appropriate.
There's more in my head that I'll show you, when there is something to show.

PR #64, opened 2015-05-11
[WIP] Timeline

C4ptainCrunch:
**  __Thoses changes heavily break the API__ **

# Timeline

In the current `master`, the `Calendar` class has a `EventList` which contains all the events. This pull requests proposes to remove `EventList` and replace it with `Timeline`

## Underlaying implementation
`Timeline` does not store any events. All events are now stored inside a  `set()`  : `Calendar.events`.
`Timeline` offers an API to get thoses events based on their date/time.

## It's advantages are : 
* Most of the time `EventList` is time-sorted but it's not guaranteed. It's seemed to be a source of confusion. Since events are stored in a  `set()` there is no more confusion. Every method of `Timeline` outputs a time-sorted iterator
* __It will be possible to support recurrent events__ : Since every method of `Timeline` are iterators, we can return recurrent events instances on the fly
* There is no more weird slicing (`Calendar.events[2014:2015:'both']`). Methods are now explicitly named
* Internal implementation of `Timeline` is far more robust (and clean) than it's ancestor

## Doc/Examples

The doc of `Timeline` is [there](http://icspy.readthedocs.org/en/timeline/api.html#timeline)

``` python
c = Calendar()
c.events # this is a simple set()
c.events.add(Event())
c.timeline # iterator over all sorted events
for event in c.timeline.on(datetime(2015, 1, 1)) # all events on 01/01/2015 (sorted of course)
    print event
 
```

---------

It is currently (partially) untested so it's not to be merged right now.
But as the number of users and contributors to `ics.py` grows, i would like to ask "permission" to break the API 

So, what do you think ?


GMLudo:
+1 for theses changes.

PR #65, opened 2015-07-09
Properties as descriptors

jammon:
Ok, this is a bigger one.

When trying to extend ics.py, I found the handling of the properties of ical components somewhat unwieldy. I thought, it should be possible, to collect all the features of the different ical property into one data structure and use this for the implementation.

So I tried to implement the as descriptors (see https://docs.python.org/2/howto/descriptor.html). Look a the new file ics/property.py to see what I mean. 

The properties of every component are stored in a dict of `ContentLine`s (`Component._properties`). Their python values are accessible as descriptor attributes to the Component objects. This makes the definition of the components nice and "DRY". Here is Event as an example:

```
class Event(Component):
    name = TextProperty('SUMMARY')
    begin = DateTimeOrDateProperty('DTSTART', validation='_validate_begin')
    end = DateTimeOrDateProperty('DTEND',
                                 validation='_validate_end',
                                 default='_default_end')
    duration = DurationProperty('DURATION',
                                validation='_validate_duration',
                                default='_default_duration')
    summary = TextProperty('SUMMARY')
    description = TextProperty('DESCRIPTION')
    location = TextProperty('LOCATION')
    url = TextProperty('URL')
    created = DateTimeProperty('DTSTAMP', default='_default_created',
                               create_on_access=True)
    uid = TextProperty('UID', default=uid_gen, create_on_access=True)
```

When implementing this I sometimes stumbled over problems with the arrow library. Since you seemed to be a little unhappy with this library as well, I removed it gradually from ics.py.

This and some other circumstances brought with them little changes to the api of ics.py:
- An Event is considered "all day", when its begin is a `date` (instead of a `datetime`). There is no such concept as "precision" any more.
- If a string representation of a date/datetime doesn't contain a time part, it is translated to a `date`. ('2015-07-09' is a date, '2015-07-09 22:00' is a datetime.)
- On all-day-Events 'end' refers to the last day of the Event. If the Event is only one day, then 'event.begin==event.end' (see RFC5545 Section 3.6.1.)
  - If an all-day-event has neither end date nor duration, it is considered to take one day ('event.begin==event.end').
- The UID and the 'created'(=DTSTAMP) attribute are generated for an Event if they don't exist when they are accessed, 
- Event.created defaults to datetime.now()
- The meaning of the modifier 'any' of `Eventlist.__getitem__` has changed, so that it selects any event that somehow overlaps the start-stop-period.
  (What is the modifier 'inc' in `Eventlist.__getitem__` meant for? I didn't understand this.)

Ok, that's it for today, there are still some rough edges left and some tests to write. But I want to give you the opportunity to comment on these ideas. The next step will be implementing the other properties and components of the ical format.


C4ptainCrunch:
I'm sorry but this PR is waaay to old and is un-mergeable so i'll close it.
It's entirely my fault, i should have reviewed it years ago. Sorry.

Issue #68, opened 2015-11-05
timezone information is dropped


ConnyOnny:
test case:

```python
from ics.parse import ContentLine
from ics.utils import iso_to_arrow
line = ContentLine.parse("DTSTART;TZID=Europe/Berlin:20151104T190000")
print(line)
arrow = iso_to_arrow(line)
print(arrow)
print(arrow.__dict__)`
```
It acts as if the given time were in UTC and completely ignores that it is in fact another time zone.

I'm digging through the code right now.

ConnyOnny:
Okay. The problem is, my failing ics-file doesn't have a VTIMEZONE definition, but still uses time zones in DTSTART etc. Then in `iso_to_arrow` the `available_tz` dict is empty and so it uses UTC.
A fix would be to first check if the given TZID parameter is a valid time zone on its own and if so just use it instead of doing a lookup in the `available_tz` dict.
I'll prepare a pull request.

btw: It's an owncloud calendar that behaves this way, so there might be many other people running into this problem…

C4ptainCrunch:
Should be fixed by #69. I'll write a test for it before closing.

PR #69, opened 2015-11-05
use TZID as time zone if possible

ConnyOnny:
fix for issue #68


C4ptainCrunch:
Awesome, thanks a lot :)

PR #70, opened 2015-11-12
make all_day more liberal

ConnyOnny:
ownCloud calendar events have an end, even if they are created as "all day". But in that case, the precision of the begin is still "day" and the timedelta between end and begin is exactly one day. So with this fix, the `all_day` property also works for calendars which have an end for an "all day" event, as long as the time delta is exactly one day and the begin precision is still "day".

### DIFF
--- a/ics/event.py
+++ b/ics/event.py
@@ -185,7 +185,7 @@ class Event(Component):
         Return:
             bool: self is an all-day event
         """
-        return self._begin_precision == 'day' and not self.has_end()
+        return self._begin_precision == 'day' and ((not self.has_end()) or self.end - self.begin == timedelta(day=1))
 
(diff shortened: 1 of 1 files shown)


ConnyOnny:
I had a typo in that PR and running the tests didn't show it :-/
I'll make a new PR.

C4ptainCrunch:
What was the problem that the test didn't show ?

ConnyOnny:
I didn't test the all_day getter enough, because `timedelta(day=1)` just fails right away, because it's called "days" and not "day". But because of short circuit boolean evaluation there was no test ever calling that code.
Also after digging through the RFC, the implemented semantics of this PR are wrong anyway, so let's just forget this PR altogether ;-)

Issue #56, opened 2014-09-14
Support badly formatted linesplits


geecko86:
_Edit_ : the file shown below is broken (it does not follow the line wrap specified in the rfc). I keep this issue open just to remember to someday add a flag to allow those illegal line-wraps and be less strict.
@C4ptainCrunch

---

Hi C4,

I haven't really dug into the specificities of the .ics format, so I apologize if I'm saying something wrong :).

Using [this](http://pastebin.com/M3Ep2peZ) as the source (which comes from [this URL](http://scientia-web.ulb.ac.be/gehol/index.php?Student/Calendar/%23SPLUS35F0ED/1-14.ics)) I've been getting errors from ics.py. The reason is that the line 649 doesn't include a ':' character.

> interrogations

```
def parse(cls, line):
    if ':' not in line:
        raise ParseError("No ':' in line '{}'".format(line))
```

I'm not sure if this is in fact allowed & your code is a bit too strict or if my source is at fault... but Google Calendar doesn't seem to have any problem with this feed.

My (quick & dirty) fix was to surround this parse() command with a try/except like this:

```
for line in unfolded_lines:
    try:
        yield ContentLine.parse(line)
    except ParseError:
        pass
```

I'm sure you can come up with something better :). This is pretty dirty so I didn't even bother making a pull request, especially since the SUMMARY field will be missing a word.

This is also one of the reasons why your [gehol-tools](https://github.com/C4ptainCrunch/gehol-tools) isn't working for me atm.

Hope this helps!

C4ptainCrunch:
I double checked, your source file is broken (not rfc compliant), you should have 2 spaces in front of line 649.
Maybe in the future we'll add an option to allow that but for now we follow the rfc ;)

(I'll edit your first post to make it clear that your source is broken)

PR #72, opened 2015-11-13
multi-day all-day events

ConnyOnny:
* events with no duration now have timedelta(0) duration instead of one second
* added tests for the new functionality
* fixed some stuff related to all-day events (e.g. the "end" getter)

Note: This doesn't fix the ownCloud all-day issue because ownCloud doesn't adhere to the RFC. For one-day all-day events, if an end date is given, it must be the same as the start date, but ownCloud gives the day after as end date. I'll file a bug report with them…


C4ptainCrunch:
Thanks a lot for this nice pull request with tests included :)
For the ownCloud problem, i'm drafting a solution to be able to activate flags to parse calendars that are not adhering to the RFC (it could also be useful to solve #56 )

PR #74, opened 2015-12-21
Use shorthand DTSTART;VALUE=DATE:YYYYMMDD for all day events

### LINKED ISSUE #73: Outputting DTSTART with VALUE=DATE
`iso_to_arrow()` supports parsing lines like `DTSTART;VALUE=DATE:20151217` but `arrow_to_iso()` always forces events to be datetime including the `Z` timezone specifier. I am running into issues supporting all day events where the appended `Z` means that I have to generate Events using the user's timezone or else it ends up being on the previous day for all timezones with negative offsets.

`DTSTART:20151215T000000Z` ends up being an all day event on Dec 14th at 10pm for America/New_York

Trii:
The current incarnation always appends a timestamp and the `Z` timezone specifier whenever you have an all day event. This means that I end up seeing events from `20151220T190000EST` until `20151220T190000EST` because of my timezone (EST -0500).

This pull request makes sure that anyone who imports the calendar has an all day event in their current timezone instead of 24 hours from UTC.

> **Value Type**
> The default value type is DATE-TIME. The time value MUST be one of the forms defined for the DATE-TIME value type. The value type can be set to a DATE value type.

Fixes #73

### DIFF
--- a/ics/event.py
+++ b/ics/event.py
@@ -18,6 +18,7 @@ from .utils import (
     iso_precision,
     get_arrow,
     arrow_to_iso,
+    arrow_date_to_iso,
     uid_gen,
     unescape_string,
     escape_string,
@@ -48,7 +49,7 @@ class Event(Component):
                  created=None,
                  location=None,
                  url=None):
-        """Instanciates a new :class:`ics.event.Event`.
+        """Instantiates a new :class:`ics.event.Event`.
 
         Args:
             name (string)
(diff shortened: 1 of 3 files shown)


C4ptainCrunch:
Hey ! Thanks for the pull request, i thought i would have to fix the issue by myself ^^
Right now, i'm quite low on time, i'll check the diff as soon as i can, sorry for the delay :/

Trii:
You're welcome! I had fixed it in my local copy to get by just in case you had your own plans. I thought I saw something about removing `arrow` and didn't know if you already had it in the works. I figured I'd submit a real PR and fix it in my fork so that I could use my fork until you get a chance to deal with this.

PR #76, opened 2016-01-13
Add TRANSP attribute support

GMLudo:
Hi,

This PR adds TRANSP attribute support: http://www.kanzaki.com/docs/ical/transp.html
I need this attribute to know if it's a free or busy event.

Thanks for your remarks.

Have a nice day.

### DIFF
--- a/ics/event.py
+++ b/ics/event.py
@@ -49,7 +49,7 @@ class Event(Component):
                  created=None,
                  location=None,
                  url=None,
-                 transparent=None):
+                 transparent=False):
         """Instantiates a new :class:`ics.event.Event`.
 
         Args:
@@ -377,8 +377,6 @@ def url(event, line):
 def transparent(event, line):
     if line:
         event.transparent = line.value == 'TRANSPARENT'
-    else:
-        event.transparent = None
 
 
 # TODO : make uid required ?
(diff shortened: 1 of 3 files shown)


C4ptainCrunch:
Hi !

Thanks for your PR :)
I think we should consider renaming the attribute because `.transp` is not very obvious for someone who hasn't read the RFC. 
Also, since the `TRANSP` prop can take only 2 values, why not cast it to a boolean ? 
I suggest naming it `.transparent` with `TRANSPARENT` casted to `True` and `OPAQUE` to `False` with a default value of `False` (or we could do it the other way around with a property named `opaque` but it might be misleading for the ones that have read the RFC)

What do you think ?

GMLudo:
Hi @C4ptainCrunch,

Your suggestions make sense, I've made a new commit based on your idea.
Is it ok to you ?

Have a nice day.

GMLudo:
Gentle ping ;-)

C4ptainCrunch:
Sorry, didn't have time to check before, i had an examination today /o\

The `TRANSP` prop has a default value, why should we bother with a `None` ?

GMLudo:
Last remark implemented.

Ok to you ?

C4ptainCrunch:
Looks great, thank you very much !

C4ptainCrunch:
Do you need this feature right now ? I could do a release on pypi if you want to ;)
Or we could wait for #64 to be merged but it's a big piece so it might take a while...

GMLudo:
I'm not in emergency, for now, I use directly my branch via pip install -e git+... ;-)

PR #83, opened 2016-06-04
Adds the ability to specify `default` parameters

PrashntS:
This PR adds the ability to specify the default parameters in `@Calendar._extracts` decorator.
If the `required` param is specified but not found, it allows to gracefully assume a default value.

#### Why?
Because I have encountered that a few iCal files have missing `VERSION` key. Rather than raising a TypeError, it makes more sense to allow this.

I'm not sure if this should be a feature, though -- since I only have one anecdotal evidence, so feel free to discuss.

Thanks!

### DIFF
--- a/ics/component.py
+++ b/ics/component.py
@@ -13,7 +13,7 @@ from .utils import get_lines
 
 Extractor = namedtuple(
     'Extractor',
-    ['function', 'type', 'required', 'multiple']
+    ['function', 'type', 'required', 'multiple', 'default']
 )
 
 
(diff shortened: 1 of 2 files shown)


C4ptainCrunch:
I looks like a nice way to parse invalid (by the rfc) calendars without scarifying the correctness of our code, i like it :)

PrashntS:
Sure, it'd be easy to do so here [0].

I've been using `ics.py` for parsing quite a few `icals` from various sources, and have seen only one instance where the `VERSION` parameter was omitted. Though, It is also true that in such cases, there's nothing that the user can do about invalid `icals`. Perhaps, something like `INFO: Version header missing, defaulting to 2.0`, would be more useful than throwing an exception but also not silently tolerating input.

[0] https://github.com/C4ptainCrunch/ics.py/pull/83/files#diff-23909cb717da4909ac47419eae4a4603R43

PR #85, opened 2016-06-17
make parser work with tabbed whitespaces

mrmadcow:
not sure whether this is rfc-compliant at all, but I have an ics file that uses tabs as line continuation characters (whitespace at beginning of line) and this breaks parsing. If attached change doesnt break other things I would like to have it added!


mrmadcow:
line[0].isspace()...

C4ptainCrunch:
Same as #86, i'll have to check if it's rfc compliant and if not, we should output a warning at least :)
It would also be nice to have a test for the pr :)

Thanks !

mrmadcow:
Is this what you requested @C4ptainCrunch ?

dgouldin:
Any reason why this couldn't be merged as-is?

aureooms:
@C4ptainCrunch referenced #86 in which @pcoueffin says that the RFC only allows a single leading space or tab to continue a line. This pull request currently allows all types of whitespace. Can someone site the RFC about this?

pcoueffin:
https://tools.ietf.org/html/rfc5545#section-3.1

Lines of text SHOULD NOT be longer than 75 octets, excluding the line
break. Long content lines SHOULD be split into a multiple line
representations using a line "folding" technique. That is, a long line can
be split between any two characters by inserting a CRLF immediately
followed by a single linear white-space character (i.e., SPACE or HTAB).
Any sequence of CRLF followed immediately by a single linear white-space
character is ignored (i.e., removed) when processing the content type.
On 4 Jan 2017 01:24, "Aurélien Ooms" <notifications@github.com> wrote:

@C4ptainCrunch <https://github.com/C4ptainCrunch> referenced #86
<https://github.com/C4ptainCrunch/ics.py/pull/86> in which @pcoueffin
<https://github.com/pcoueffin> says that the RFC only allows a single
leading space or tab to continue a line. This pull request currently allows
all types of whitespace. Can someone site the RFC about this?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/C4ptainCrunch/ics.py/pull/85#issuecomment-270328205>,
or mute the thread

dgouldin:
Beat me to it 😉 . Looks like `isspace` is locale-dependent: https://docs.python.org/2/library/stdtypes.html#str.isspace

So the only change that might make this more rfc-compliant is something like `line[0] in (' ', '\t')`

PR #86, opened 2016-06-20
Fix for leading multiple whitespace (both spaces and tabs) on a conti…

pcoueffin:
…nued line.

### DIFF
--- a/ics/event.py
+++ b/ics/event.py
@@ -403,6 +403,11 @@ def alarms(event, lines):
 
     event.alarms = list(map(alarm_factory, lines))
 
+@Event._extracts('CATEGORIES')
+def categories(event, line):
+    if line:
+        event.categories = line.value
+
 
 # -------------------
 # ----- Outputs -----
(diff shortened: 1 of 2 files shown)


C4ptainCrunch:
I don't think tabs are valid whitespace in front of a line in an ics file (i'll check). If not, we should at least output a warning.
Could you do a separated pull request for the ´categories´ bit ? (it lacks tests and it would be nice to use a list instead)

pcoueffin:
OK.  I got a file with leading tabs from Google calendar.  Did not check
the rfc.

I'll break out the categories separately.

pcoueffin:
Checked rfc.  I got it half right...

That is, a long line can be split between any two characters by inserting a
CRLF immediately followed by a single linear white-space character (i.e.,
SPACE or HTAB).

I'll go back to not using lstrip...

d-k-c:
Any progress on this issue? I get bitten by it quite often with invitations generated by Outlook 14.0

C4ptainCrunch:
I'm closing this as the PR does not follow the RFC and i don't have any example that shows a bug in the parsing. If you do have one, please open an issue with the example attached :)

PR #89, opened 2016-09-29
Fix missing replace for \r

etnarek:


C4ptainCrunch:
@etnarek : build failed ;)

PR #93, opened 2016-10-05
Fix some bugs

wom-bat:
There are two patches in this pull request.  The first fixes all-day events, by allocating the local timezone to them (otherwise, filtering an event list by .today() may miss an all day event for today, if UTC today doesn't overlap local today); the second elides SEQUENCE lines, because the parser can;t cope with them.


wom-bat:
I'll fix the style issue and resubmit.

PR #94, opened 2016-10-05
Fix some bugs: SEQUENCE and all-day events.

wom-bat:
I've fixed the build failure (I hope!).  These fixes uncover some problems in the test suite:
the end date for an all day event is the day _after_ the end (the event ends at 0:00 that day, or midnight the previous day() (so test_all_day_duration is incorrect); and make_all_day is NOT idempotent unless its called in the timezone of the event.

So you will see two test failures.

### DIFF
--- a/ics/event.py
+++ b/ics/event.py
@@ -208,7 +208,10 @@ class Event(Component):
         """Transforms self to an all-day event.
 
         The event will span all the days from the begin to the end day.
+        If the end time is 0 (midnight the previous day),
+        the last day will not be included.
         """
+
         was_instant = self.duration == timedelta(0)
         old_end = self.end
         self._duration = None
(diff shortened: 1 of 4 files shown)


wom-bat:
I''ve now fixed make_all_day() and adjusted the test case that was failing.  The remaining travis failure appears to be a test system configuration error.

C4ptainCrunch:
PR #144 might be related

N-Coder:
Hey! Implementing `make_all_day` properly seems to be a pretty complicated thing, see [here](https://github.com/C4ptainCrunch/ics.py/pull/222#issuecomment-583764888) for details. That's the reason why I started the implementation for the next version 0.8 entirely focused on correctness (see details in the linked PR). Hopefully, having that PR merged has now fixed the all-day issues and also the local timezone issue addressed in this PR. For the SEQUENCE property I guess that it should be okay if they just get added into the `extra` Container until we can handle them properly. So I guess we can close this PR, if that's okay for you. If I missed something, just comment here. Thanks for you contribution! :)

PR #105, opened 2017-09-06
Error parsing events dates in Arrow 0.10.0

Ni-Knight:
For some reason this broke with the new Arrow. adding the format as below seems to do the trick.

### DIFF
--- a/ics/utils.py
+++ b/ics/utils.py
@@ -45,13 +45,13 @@ def iso_to_arrow(time_container, available_tz={}):
         val = time_container.value
 
     if tz and not (val[-1].upper() == 'Z'):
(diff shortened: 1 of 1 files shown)


Ni-Knight:
Err, my bad seems like its the UTC `Z` sign which drives him crazy so with this change it just ignores TZ if not UTC.

C4ptainCrunch:
Fixing this issue is much more complicated as the format is not always `YYYYMMDDTHHmmss`.
I'll close the PR for now as we need a more robust solution (and maybe totaly remove arrow and write our own parsing functions as the arrow API changed quite a lot)

PR #113, opened 2018-02-19
Todo support

tgamauf:
Hi,

thank's for the library! Since VTODO support was missing and I needed it, I have implemented it. The code is covered by unit tests and I also have tested it with my own ics todo lists.

Beside the mandatory attributes (dtstamp, uid), I've also implemented the attributes completed, created, description, dtstart, location, percent, priority, summary, and url.
I tried to keep my implementation of Todo close to your implementation of Events, but I diverged in a few areas:

1. Attribute DTSTAMP in Events is mapped to variable created. As the standard defines a CREATED attribute, I didn't follow you there, but added a separate variable dtstamp. created is now filled from CREATED.
2. I thought about implementing the logical operators (__xor__, __or__, __and__, ...), join, and so on, but decided against it. The reason is that I just don't see how this really would make sense for todo list entries.
3. The comparison operators (<, <=, =>, >) operate on the due time (DUE), not the begin time (DTSTART), as I believe that this makes more sense. At least this is the case for the way I use them.

In addition, I found some behaviors of the parser I wasn't sure about, but didn't investigate:

1. A timestamp which contains seconds, e.g. 2018-02-19 12:00:**05** will be stored as 2018-02-19 12:00:**00**. In my opinion seconds are irrelevant here, but it was surprising none-the-less.
2. Timezone information of time fields is ignored, e.g. if COMPLETED:;TZID=Europe/Brussels:20180418T150001 is parsed, it won't store the timezone.

It would be amazing if you could have a look and if it makes sense merge it into your library!

Thanks,
Thomas


eega:
I didn't consider Python2.7 - I will fix it and recommit as soon as possible.

C4ptainCrunch:
Wow, that's a pretty big PR, thanks a lot !
TODO support was long overdue (pun intended), thank you for writing it :)

It might need a little more documentation but i'll merge anyway, documentation can come later.

--------------

> I thought about implementing the logical operators (xor, or, and, ...), join, and so on, but decided against it. The reason is that I just don't see how this really would make sense for todo list entries.

:+1: 

> The comparison operators (<, <=, =>, >) operate on the due time (DUE), not the begin time (DTSTART), as I believe that this makes more sense. At least this is the case for the way I use them.

It makes sense for me too.

> A timestamp which contains seconds, e.g. 2018-02-19 12:00:05 will be stored as 2018-02-19 12:00:00. In my opinion seconds are irrelevant here, but it was surprising none-the-less.

It look awfully like a bug :( Could you file an issue if you have some spare time ?

> Timezone information of time fields is ignored, e.g. if COMPLETED:;TZID=Europe/Brussels:20180418T150001 is parsed, it won't store the timezone.

This also looks like a bug, it might be from all the arrow mess that i have to clean. I'll look into it (but feel free to open an issue with code example if you want to)

PR #116, opened 2018-03-08
Reference paramstr instead of line in ParseError

jlyoung:
ParseError is raised when '=' is not in paramstr.
But the error message that's returned says '=' is not in line (not paramstr).


C4ptainCrunch:
Thanks !

PR #123, opened 2018-04-28
feature: implement event geo

johnnoone:


C4ptainCrunch:
Hi !

Thank you for submitting a PR :)
It seems that you still have some problems in the unit tests, could you fix them ?

I would prefer the full length `latitude` and `longitude` instead of `lat` and `lon`, `lng`, `long`. Could you change that ?

Also, it would be great if you could write a little bit of documentation :)

pedromorgan:
Please `lat,lon`, not full length!

C4ptainCrunch:
I understand that you don't want to type extra characters but everybody has an autocomplete nowadays.

Besides that, you just made my point, the PR uses `lng` and you asked for `lon` : there is only one way to write the long form but many ways to write the short form so people make mistakes with the short form.

pedromorgan:
I don't care what the short version is, its much easier that the long version !

PR #124, opened 2018-04-28
feature: implement event and todo status

johnnoone:


C4ptainCrunch:
It seems that you are mixing the `PARTSTAT` and `STATUS` properties, am i right ?

PR #138, opened 2018-07-09
Add attendees in events

t00n:
* params RSVP and CN
 * value "mailto:<email>"


ics/event.py
```diff
@@ -523,6 +533,17 @@ def categories(event, line):
             event.categories.update({unescape_string(cat)})
 
 
+@Event._extracts('ATTENDEE')
+def attendee(event, line):
+    if line:
+        # skip the "mailto:" part to get only email
```
C4ptainCrunch: an `assert line.value[:7] == "mailto:"` might be useful to avoid eating garbage if the input is invalid. 

> fail early, fail often.

PR #139, opened 2018-07-28
Upgrade Arrow to 0.10.0

### LINKED ISSUE #77: Upgrade Arrow to 0.10.0
I see you are using an old version of Arrow (0.4.2). Now that `datetime.date` instances can be used for all day events, the version of Arrow you pin to doesn't support `arrow.get(datetime.date)` so the setter for `Event.begin` raises an exception. Is there a reason why it hasn't been upgraded yet? I can take a look if not.

```
  File "/Users/Josh/Developer/Github/training-cal/src/lib/ics/event.py", line 81, in __init__
    self.begin = begin
  File "/Users/Josh/Developer/Github/training-cal/src/lib/ics/event.py", line 112, in begin
    value = get_arrow(value)
  File "/Users/Josh/Developer/Github/training-cal/src/lib/ics/utils.py", line 157, in get_arrow
    return arrow.get(value)
  File "/Users/Josh/Developer/Github/training-cal/src/lib/arrow/api.py", line 23, in get
    return _factory.get(*args, **kwargs)
  File "/Users/Josh/Developer/Github/training-cal/src/lib/arrow/factory.py", line 146, in get
    raise TypeError('Can\'t parse single argument type of \'{0}\''.format(type(arg)))
TypeError: Can't parse single argument type of '<type 'datetime.date'>'
```

muffl0n:
Applies patch from https://github.com/C4ptainCrunch/ics.py/issues/77 (closes #77 )

### DIFF
--- a/ics/event.py
+++ b/ics/event.py
@@ -168,7 +168,7 @@ class Event(Component):
     @end.setter
     def end(self, value):
         value = get_arrow(value)
-        if value and value < self._begin:
+        if value and self._begin and value < self._begin:
             raise ValueError('End must be after begin')
 
         self._end_time = value
--- a/ics/todo.py
+++ b/ics/todo.py
@@ -179,7 +179,7 @@ class Todo(Component):
     @due.setter
     def due(self, value):
         value = get_arrow(value)
-        if value and value < self._begin:
+        if value and self._begin and value < self._begin:
             raise ValueError('Due must be after begin')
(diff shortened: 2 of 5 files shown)


C4ptainCrunch:
Thanks for submitting a PR to ics.py :) However, it seems to break the tests. Could you look into it ? :)

PR #140, opened 2018-08-05
Update README and .gitignore

aureooms:
Explain how to setup `virtualenv` for n00bs like me and ignore related files.


C4ptainCrunch:
Actually, it's better to use `virtualenv ve/` so all the `bin`/`lib`/`include` are in a directory (i use `ve` but `venv` or other variants are valid too. We could ignore all of them)

Issue #127, opened 2018-05-21
ACTION:NONE Invalid alarm action


aureooms:
I got a calendar (google) with the following type of alarms. Does not look like it is in the RFC.

    BEGIN:VALARM
    ACTION:NONE
    ...
    END:VALARM

Current code throws an Exception with message `Invalid alarm action`.

MHohenberg:
Just to put my hat into the ring: I have this issue as well. 

This is not a 'feature request', IMHO, but a showstopper-level bug.

aureooms:
We could have options to choose the RFC-compliance strictness level. The checking part could be a separate plugin.

balzer82:
As a workaround I simply solved it by manipulating the ics string like this:

```python
ics = requests.get(url).text
ics = ics.replace('BEGIN:VALARM\r\nACTION:NONE','BEGIN:VALARM\r\nACTION:DISPLAY\r\nDESCRIPTION:')
c = Calendar(ics)
```

n2o:
I am having the same issue when reading a Nextcloud Calendar and a `ACTION:EMAIL` description in it.

PR #143, opened 2018-08-10
Support VALARM of ACTION:NONE

zagnut007:
It looks like Google ICS files may include a VALARM action of None which throws an error. Although this type isn't mentioned in the RFC that I could find, I think ignoring these alarms is valid based on the action type and should throw an error given the popularity of Google Calendar.

This addresses: https://github.com/C4ptainCrunch/ics.py/issues/127


C4ptainCrunch:
Thanks !
(closes #127)

PR #144, opened 2018-09-27
Fix all-day events lasting multiple days

raspbeguy:
Before this patch, all-day that lasted multiple days had a wrong `DTEND` field: it kept hour/minute/second information, which is a problem for some calendar client like khal and ICSdroid.

This patch fixes this issue.

### DIFF
--- a/ics/event.py
+++ b/ics/event.py
@@ -547,7 +547,9 @@ def o_all_day(event, container):
     if event.begin and event.all_day:
         container.append(ContentLine('DTSTART', params={'VALUE': ('DATE',)},
                                      value=arrow_date_to_iso(event.begin)))
-
+        if event._end_time:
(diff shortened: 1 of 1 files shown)


C4ptainCrunch:
Thanks a lot ! I added your contribution to the CHANGELOG so i had to make a new branch (and thus a new PR, #152) but your name and commits are preserved :)

raspbeguy:
You could have just pushed the modification in the readme into this branch, so the commit would have stayed mine, but I guees this will do too.
Thanks.

C4ptainCrunch:
I did not realize that when squashing the PR as i was the user that created the PR i would get the commit. So i amended the commit and you are the author again :)
https://github.com/C4ptainCrunch/ics.py/commit/40a1541260a8e22d758f60d7c43d424fc99dc1f0
Sorry for the first bad commit (and also for the force push on master to fix it :D)

PR #145, opened 2018-10-03
requirements: Lock arrow version to >0.4.2

chauffer:
Tested with 0.11.2. Necessary due to granularity parameter in
humanize()


chauffer:
@C4ptainCrunch is this project maintained? there're plenty of unanswered MRs

C4ptainCrunch:
@chauffer The project is somewhat maintained, it depends on my free time unfortunately ...

Thanks for you PR :) Upgrading arrow was on my todo list for a long time.
Unfortunately the tests do not pass, it looks like it's due to the hardcoded value in `arrow.get` that does not accept enough formats. Could you try to fix that by allowing all datetime formats described in the RFC ?

Issue #155, opened 2018-12-11
make_all_day() issues


perette:
Support for all-day events is ugly and confusing.

`make_all_day` treats an event's begin/end times as an inclusive date range, and converts it to half-open.  For `begin`, it's set to the start of its day, but for `end` it's bumped to the start of the next day, even if it's already at the start of a day.  Additionally, if the begin time is subsequently changed, the event unexpectedly reverts to a not-full-day event.  This is not documented and has a side effect that calling `make_all_day` again causes the end to be bumped up by one, as it is re-converted from inclusive to half-open range.  However, setting the end time does not effect the all-day status, nor is the value rounded or bumped; you can make an all-day event that's 1 day, 3 hours and 5 minutes.

I'd like to suggest a combination of behavioral and documentation improvements.  I'm willing to commit time to make these happen, but want to gather input and confirm I'm sane in my approach:

Definite code changes:
- If an event has a *duration* instead of an end time, `make_all_day` should round the duration up to the next 24-hour interval.  If it's already an even day, the duration should be unchanged.  (Currently, a duration is converted to a fixed end-time, subject to bumping on successive `make_all_day`.)

Definite documentation changes:
- Describe how how start/end times/dates are transformed by `make_all_day`.

Possible code changes:
- Instead of having `begin`'s setter implicitly change events to non-all-day event, instead have it round the provided begin time to the start of day when it's an all-day event.
- Modify `end`'s setter to also round, but round up rather than down.  Setting to start of day should not increment to next day (to retain compatibility with current behavior), but that should be clearly documented.
- Add a boolean parameter ala `make_all_day(ad = True)`, with default True which would make the event all-day.  If the parameter is false, revert to a non-all-day event.

That seems sane to me, as it's changing already-blurry areas for behavior.  But it's still changing behavior and risks breaking code using the library.  Alternately, there's the documentation route:
- Document that the event loses it's all-day property if the begin time is set.
- Indicate that `make_all_day()` can be called to restore all-day status, but doing so will bump the end date again.  Assuming the change for all-day durations is in place, suggest using duration instead of end date to avoid bumping.
 
Thoughts?

C4ptainCrunch:
Hi @perette !

Half a year later, i'm answering to your post, sorry about the delay :/
Thank you for the well constructed and augmented post.

Changing the behaviour looks to me the way to go, after all, we are not yet in 1.0, breaking changes can and will happen :)

N-Coder:
[...]

PR #159, opened 2018-12-23
Improve all-day event handling.  See issue #155, _make_all_day() issues.

perette:
This patch includes the set of changes for all-day events proposed in issue #155, except instead of adding a parameter to `make_all_day()` to revert to non-all-day events, I added a second member function, `cancel_all_day()`.  Corresponding unit tests are included.

### DIFF
--- a/ics/event.py
+++ b/ics/event.py
@@ -124,16 +124,20 @@ class Event(Component):
         |  May be set to anything that :func:`Arrow.get` understands.
         |  If an end is defined (not a duration), .begin must not
             be set to a superior value.
+        |  For all-day events, the time is truncated to midnight when set.
         """
         return self._begin
 
     @begin.setter
     def begin(self, value):
         value = get_arrow(value)
+        precision = 'day' if self.all_day else 'second'
+        if value is not None:
+            value = value.floor(precision)
         if value and self._end_time and value > self._end_time:
             raise ValueError('Begin must be before end')
         self._begin = value
-        self._begin_precision = 'second'
+        self._begin_precision = precision
 
     @property
     def end(self):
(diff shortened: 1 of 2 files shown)


ics/event.py
```diff
@@ -217,27 +242,49 @@ def all_day(self):
         # the event may have an end, also given in 'day' precision
         return self._begin_precision == 'day'
 
+    def cancel_all_day(self):
```
C4ptainCrunch: I'm not a fan of the naming, maybe doing `make_all_day(False)` like you suggested could be better

tests/event.py
```diff
@@ -1,3 +1,5 @@
+# vim: ts=4:sw=4:expandtab
```
C4ptainCrunch: Could you remove this ? :)

tests/event.py
```diff
@@ -81,6 +84,95 @@ def test_make_all_day2(self):
...
+
+        self.assertEqual(e._begin, begin)
+        self.assertEqual(e.begin, begin)
+
+        self.assertEqual(e._end_time, end)
+        self.assertEqual(e.end, end)
+
+        e.make_all_day()
+
+        self.assertEqual(e._begin, arrow.get ("2018-12-23"))
```
C4ptainCrunch:
```suggestion
        self.assertEqual(e._begin, arrow.get("2018-12-23"))
```

tests/event.py
```diff
@@ -81,6 +84,95 @@ def test_make_all_day2(self):
...
+        self.assertEqual(e.begin, arrow.get ("2018-12-23"))
```
C4ptainCrunch:
```suggestion
        self.assertEqual(e.begin, arrow.get("2018-12-23"))
```

perette:
I'm presently traveling, but plan to make the requested changes when I'm with a real machine instead of a toyblet.  ETA late June/early July/2-3 weeks.

C4ptainCrunch:
I did not notice you pushed the changes i asked, sorry !
I fixed some conflicts and pushed to master in https://github.com/C4ptainCrunch/ics.py/commit/ed14cde096ad5b50bfa768c1fd32892eba744087. Your changes will be in 0.7 !

PR #163, opened 2019-01-02
Make Alarm hashable

### LINKED ISSUE #130: Adding alarms to Event Objects
Unless I'm mistaken, it is not possible to pass an Alarm to Event seeing as how the Event.alarms is a set and Alarm Objects are unhashable. 
```python
class Event(Component):
    def __init__(self, ...):
        ...
        self.alarms = set()
        ...
        if alarms is not None:
            self.alarms.update(set(alarms))
```

Furthermore, in **ics.py/tests/alarm**, the `test_alarm_without_repeat_extraction` function treats the alarms of Event as a list (in the sense of how it accesses the first alarm):
```python
    def test_alarm_without_repeat_extraction(self):
        c = Calendar(cal21)
        a = c.events[0].alarms[0]
        ...
```

Philiptpp:
Add __hash__ method to the Alarm class to make it hashable
in order to be able to add an Alarm list to the Events.alarms set.

Fixes #130

### DIFF
--- a/ics/alarm.py
+++ b/ics/alarm.py
@@ -157,6 +157,9 @@ class Alarm(Component):
 
         return '<{0}>'.format(value)
 
+    def __hash__(self):
+        return hash(repr(self))
+
     def __ne__(self, other):
         return not self.__eq__(other)
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -18,6 +18,7 @@ Other contributors, listed alphabetically, are:
 * `@jammon <https://github.com/jammon>`_
 * `@mrmadcow <https://github.com/mrmadcow>`_
 * `@perette <https://github.com/perette>`_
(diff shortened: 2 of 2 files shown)


ics/alarm.py
```diff
@@ -157,6 +157,9 @@ def __repr__(self):
 
         return '<{0}>'.format(value)
 
+    def __hash__(self):
```
Philiptpp: Added the hash method to the Alarm class to be able to add a list of Alarm to the Event.alarm set

C4ptainCrunch:
Hi @Philiptpp 

Sorry for the delay, it's hard to keep maintaining this project in my free time.
Could you just add a test proving that the issue is fixed ? (or [give me access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to the branch and i'll do it myself)

Thank you !

PR #168, opened 2019-01-13
Add priority, due date and rrule to Event class 

btrevizan:
rrule is totally functional and is recognise by Timeline methods.

Priority and due date could be useful for some applications. The existence of such properties doesn't affect the package, since it is ignored by the parser if "PRIORITY" and "DTDUE" are not specified in .ics or.ical files.

### DIFF
--- a/ics/event.py
+++ b/ics/event.py
@@ -484,9 +484,9 @@ class Event(Component):
         for event in self.repeat():
             # if start is between the bonds and stop is between the bonds
             if start <= event.begin <= stop and start <= event.end <= stop:
-                return True
+                return event
 
-        return False
+        return None
 
     def overlaps(self, start, stop):
         """Return True if event overlaps start-stop interval.
@@ -499,9 +499,9 @@ class Event(Component):
         for event in self.repeat():
             # if start is between the bonds or stop is between the bonds or event is a superset of [start,stop]
             if (start <= event.begin <= stop or start <= event.end <= stop) or event.begin <= start and event.end >= stop:
-                return True
+                return event
 
-        return False
+        return None
 
     def starts_after(self, instant):
         """Return True if event starts after instant.
@@ -512,9 +512,9 @@ class Event(Component):
         """
         for event in self.repeat():
             if event.begin > instant:
-                return True
+                return event
 
-        return False
+        return None
 
     def occurs_at(self, instant):
         """Return True if event occurs at instant.
@@ -525,9 +525,9 @@ class Event(Component):
         """
         for event in self.repeat():
             if event.begin <= instant <= event.end:
-                return True
+                return event
 
-        return False
+        return None
 
     def repeat(self):
         """Iterate over the events according with rrule.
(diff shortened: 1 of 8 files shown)


N-Coder:
Hey! As already said in a few places (see [here](https://github.com/C4ptainCrunch/ics.py/pull/222#issuecomment-583764888) or [here](https://github.com/C4ptainCrunch/ics.py/pull/180#issuecomment-602190918)), I'm currently working on the next version of ics.py and RRULEs are something I'd love to finally have proper support for. So, first of all, a big thank you for your contribution! Especially interpreting the rule and generating the next occurrences was probably a lot of work. The only thing I'm wondering about is whether we could build upon the [rrule functionality of dateutil](https://dateutil.readthedocs.io/en/stable/rrule.html) (which is already a dependency of ics.py) to reduce complexity a little on our side. I didn't read into the details of RRULEs yet, so it would be great to hear your opinion in this case!

btrevizan:
Hey. Thank you! The [rrule functionality of dateutil](https://dateutil.readthedocs.io/en/stable/rrule.html) has already all the functions needed. I can't imagine a scenario where you will need more than what it is already implemented, but nonetheless if you need to implement helper functions it wouldn't be too much work. I encourage the use of dateutil for rrules.

pauladams8:
I could use `rrule` support, is there any chance this PR could be merged soon?

ics/event.py
```diff
@@ -422,12 +467,184 @@ def clone(self):
...
+        new_event.uid = uid_gen()
+        return new_event
+
     def __hash__(self):
         """
         Returns:
             int: hash of self. Based on self.uid."""
         return int(''.join(map(lambda x: '%.3d' % ord(x), self.uid)))
 
+    def is_included(self, start, stop):
```
pauladams8, line 485: within? (better English)

ics/repeatable.py
```diff
@@ -0,0 +1,273 @@
...
+                 freq="DAILY",
+                 until=None,
+                 count=None,
+                 interval=1,
+                 byday=[],
+                 bymonthday=[],
+                 bymonth=[]):
+        """Instantiate a new Repeatable rule.
+
+        :param freq: str
```
pauladams8, line 33: Could we use a `datetime.timedelta` to derive the frequency, rather than a hardcoded string? E.g. `timedelta(weeks=1)` rather than `"WEEKLY"`

pauladams8:
Looks like this will also need to be updated to replace `arrow` with the native `datetime` as `arrow` has been removed in #222

ics/repeatable.py
pauladams8: The `timedelta` could be seralized later into an RFC compatible frequency

N-Coder:
Hey, PR #249 adds `rrule` support based on the `dateutil` library using the new serialization infrastructure, superseding this PR. The initial goal was to enable correct handling of vTimezones (see e.g. [here](https://github.com/C4ptainCrunch/ics.py/blob/e404f167d97461ff1915fa727ff9e3d1e56dc11a/doc/timezone.rst#converting-to-the-builtin-vtimezone-definition) for a working example), but that should also be directly transferable to Events. This PR is still open because I still need to cherry-pick the recurrence handling for the query operations of Event and Timeline into the other PR.

PR #249 still needs some polish, but we should be able to merge that soonish. After that, the only thing that is in the way of a v0.8 alpha release is probably some possibility to tune down the parser strictness, as the new parser implementation (which was focussed on also being able to validate files) sometimes seems to nit-picky to handle practically real-world data with it.

asimonson1125:
Anything we can do to help to add rrule support?  I'd very much appreciate this

N-Coder:
Support for parsing `rrule`s is already merged, so you could try adding an `rruleset` field to the `CalendarEntryAttrs` class (so that both `Event` and `ToDo` get that functionality). The harder part is then correctly interpreting repetitions, especially when instantiating a repeating event or when parsing a recurring event with modifications to individual instances, see [this](https://www.nylas.com/blog/calendar-events-rrules/) or [this](https://icalevents.com/4437-correct-handling-of-uid-recurrence-id-sequence/) article for an introduction on this complexity. Fortunately, there is at least some [prior work](https://pypi.org/project/recurring-ical-events/), but we probably should also check how the big Calendaring solutions (Google, Apple, Thunderbird, Evolution and probably breaking everything, Outlook) do this. The big issue then is to design a user-facing API that helps with
- creating modified instances of repeating events, also including cancellations
- linking a modified instance back to its original instance, where the modication was possibly done by some other software
- listing all the actual occurences of a recurring event, taking modified instances and cancellations into account
  (here we also need to think about how to represent recurrence instances that haven't been modified, i.e. won't become separate objects in the ics file)

As far as I'm aware, this PR only tackled the last point while ignoring modified instances. So copying the respective changes over might be helpful, but there still is a lot of work to do here, but maybe we can reuse parts of the recurring-ical-events project (depending on the extent of this, we might need to be aware that this pulls in [LGPLv3 code](https://dwheeler.com/essays/floss-license-slide.html), but that should be fixable), and their testsuite should also be very useful. So maybe try looking at what [RFC 5545](https://tools.ietf.org/html/rfc5545) officially mandates, how [recurring-ical-events](https://pypi.org/project/recurring-ical-events/) solved that and what other Calendaring solutions do, and try to come up with an API that tackles the 3 points listed above.

C4ptainCrunch:
I'm doing a bit of cleanup. This PR seems stale and not ready to be merged so i'll close it but if somebody wants to work on it again, feel free to reopen !

PR #172, opened 2019-03-23
Feature: Last modified attribute

Timic3:
This pull request adds last modified attribute, which is very useful, if you want to check if/when the event has changed.

I also removed meta symlink, because it was outputting import errors, I hope that's okay.


C4ptainCrunch:
Great! Thank you :)

PR #174, opened 2019-04-05
don't convert timezone of dates, fixes #173

### LINKED ISSUE #173: all-day event is still off by one day
This is another all-day event off-by-one error, related to #92 and #150, but not fixed by bd918ec7453a7cf73a906cdcc78bd88eb4bab71b.
Let's say the begin time of a not-all-day event is in a timezone east of UTC (having a positive offset), for example `2019-05-29T12:34:56+02:00`.
We now convert this Event to an all-day event using `make_all_day`, which will internally call `floor('day')` on that timestamp. As Arrow will still keep the timezone of the initial timestamp, the result is `2019-05-29T00:00:00+02:00`, midnight on the beginning of 29th of May at Central European Summer Time.
When this event is now serialized to its String/ics representation, the function [`arrow_date_to_iso`](https://github.com/C4ptainCrunch/ics.py/blob/fb7da6c1c1a0e2470abe92af1726aa7483d38767/ics/utils.py#L190) will be used to format this timestamp as date string. This function will convert the timestamp to UTC using `.astimezone(tzutc)`, yielding `2019-05-28 22:00:00+00:00` - which is on the previous day.
The string that is finally written to the output is - you probably guessed it -  `DTSTART;VALUE=DATE:20190528`.

This could probably be fixed by simply dropping the `.astimezone(tzutc)` call in the case of `arrow_date_to_iso`, as we are interested in the date per se not the exact time(-zone), but I'm not aware whether that would have any side-effects. Could you please fix this before making a new release (as per #153)?

Keep up the good work, this library is otherwise very pleasant to use (and I like the approach using extracts/outputs annotations).

N-Coder:
fixes #173, see description there

### DIFF
--- a/tests/event.py
+++ b/tests/event.py
@@ -339,13 +339,15 @@ class TestEvent(unittest.TestCase):
         """
         for tz in range(-3, 3):
             for hour in range(-3, 3):
-                e = Event(begin=Arrow(2019, 05, 29, tzinfo="%+02d:00" % tz).shift(hours=hour))
-                e.make_all_day()
-                if hour < 0:
-                    self.assertIn('DTSTART;VALUE=DATE:20190528', str(e).splitlines())
-                else:
-                    self.assertIn('DTSTART;VALUE=DATE:20190529', str(e).splitlines())
-        
+                for duration in range(0, 4):
+                    start = arrow.Arrow(2019, 5, 29, tzinfo="%+03d:00" % tz).shift(hours=hour)
+                    e = Event(begin=start, duration=td(days=duration))
+                    e.make_all_day()
+                    lines = str(e).splitlines()
+                    self.assertEqual(e.duration.days, duration)
+                    self.assertIn('DTSTART;VALUE=DATE:%s' % start.format('YYYYMMDD'), lines)
+                    self.assertIn('DTEND;VALUE=DATE:%s' % start.shift(days=duration).format('YYYYMMDD'), lines)
+
     def test_transparent_input(self):
         c = Calendar(cal19)
         e = next(iter(c.events))
--- a/tests/event.py
+++ b/tests/event.py
@@ -333,6 +333,19 @@ class TestEvent(unittest.TestCase):
         # no time or tz specifier
         self.assertIn('DTSTART;VALUE=DATE:20151221', str(e).splitlines())
 
+    def test_all_day_tz_independent(self):
+        """
+        see Issue #173
+        """
+        for tz in range(-3, 3):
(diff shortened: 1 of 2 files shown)


C4ptainCrunch:
Hi @N-Coder !

Thank you for your PR :) Could i just ask you to add tests to prove the issue is fixed ?

tests/event.py
```diff
@@ -333,6 +333,19 @@ def test_all_day_outputs_dtstart_value_date(self):
         # no time or tz specifier
         self.assertIn('DTSTART;VALUE=DATE:20151221', str(e).splitlines())
 
+    def test_all_day_tz_independent(self):
+        """
+        see Issue #173
+        """
+        for tz in range(-3, 3):
+            for hour in range(-3, 3):
+                e = Event(begin=Arrow(2019, 05, 29, tzinfo="%+02d:00" % tz).shift(hours=hour))
```
C4ptainCrunch:
```suggestion
                e = Event(begin=Arrow(2019, 5, 29, tzinfo="%+02d:00" % tz).shift(hours=hour))
```
Else it would be considered as octal :)

N-Coder:
I'm now also testing for the correct end time of all day events and I think I found another oddity: when you create an event with a certain duration in days and convert it to an all-day event, it suddenly gets longer by one day. If you factor in timezones, start times that don't equal midnight and durations that only roughly equal one day, it get's even weirder:
https://gist.github.com/N-Coder/28c2b949a3fb1793b1433915c8ab4686

This is caused by the `make_all_day` logic:
```
begin_day = self.begin.floor('day')
end_day = self.end.floor('day')
self._begin = begin_day
# for a one day event, we don't need a _end_time
if begin_day == end_day:
    self._end_time = None
else:
    self._end_time = end_day + timedelta(days=1)
self._duration = None
```

One possible solution would be using `end_day = self.end.`**`ceil`**`('day')` (which doesn't change the value if it equals midnight) instead of always blindly adding one day. Another - in my eyes more logical - option would be completely ignoring the end time, but rounding the initial duration to days (using `floor/round/ceil`, which one to use according to the principle of least surprise is probably up to discussion) and then adding this amount of days back again after the conversion.

N-Coder:
I just found @perette's PR #159 and issue #155, which seem to go in a similar direction as my comments above. Maybe we should finish those points first an then rebase this PR on top of #159.

N-Coder:
I guess we could've also merged this PR for 0.7, but I kind of forgot and it's to late now. So nevermind, I'm closing this.

PR #178, opened 2019-06-02
Add support for classification in event

### LINKED ISSUE #177: Add CLASS support
Are there any plans to support the CLASS Element?
I need the following values:
* CLASS:PRIVATE
* CLASS:PUBLIC

PascalBru:
Now support also classification in events.
See https://tools.ietf.org/html/rfc5545#section-3.8.1.3 for details
Closes #177


tests/event.py
```diff
@@ -457,3 +457,31 @@ def test_issue_92(self):
...
+        e = Event(name="Name", classification='PRIVATE')
+        self.assertIn("CLASS:PRIVATE", str(e).splitlines())
+
+    def test_classification_bool(self):
+        with pytest.raises(ValueError):
+            Event(name="Name", classification=True)
+
+    def test_classification_wrong_value(self):
+        with pytest.raises(ValueError):
+            Event(name="Name", classification='wrong_value')
```
aureooms: 1. Missing tests for classification not set (maybe anti-pattern or do we want `e.classification` to be `None`)?
  2. Missing tests for other valid values (`CONFIDENTIAL`, `iana-token`, `x-name`).

ics/event.py
```diff
@@ -71,6 +72,7 @@ def __init__(self,
             alarms (:class:`ics.alarm.Alarm`)
             categories (set of string)
             status (string)
+            classification (string)
```
aureooms: Is it defined as `classification` or `class` in the RFC?

aureooms: Maybe clashes with python property `class`?

aureooms: OK, more like a syntax error:

    >>> x.class
      File "<stdin>", line 1
        x.class
              ^
    SyntaxError: invalid syntax

PascalBru: In the RFC they call it `classification` and the name of the Property is `CLASS`, so they use both names as you can see here https://tools.ietf.org/html/rfc5545#section-3.8.1.3

when did you get the syntax error?

tests/event.py
PascalBru: 1. All existing test doesn't have a value for classification, there fore I thought no extra test is needed and in the RFC it is written
> If not specified in a component that allows this property, the default value is PUBLIC

2. missing tests are added with commit ab8dc7ffd21221a5bb0b8591f216a44dec55bfa1

ics/event.py
aureooms: Ok. I meant using `class` instead of `classification` would not work.

PascalBru: Okay, now in understand.
Hope everything is fine now

tests/event.py
aureooms: Thanks for the added tests.
Should still decide whether `e.classification` should return `"PUBLIC"`, `None`, or raise an exception.
@C4ptainCrunch Care to chime in?

ics/event.py
```diff
@@ -253,6 +256,22 @@ def status(self, value):
...
         self._status = value
 
+    @property
+    def classification(self):
+        return self._classification
+
+    @classification.setter
+    def classification(self, value):
+        if value is not None:
+            classifications = ('PUBLIC', 'PRIVATE', 'CONFIDENTIAL', 'iana-token', 'x-name')
```
C4ptainCrunch: iana-token and x-name are not strings but references to other parts of the spec. They are defined here : https://tools.ietf.org/html/rfc5545#section-3.1
```
iana-token    = 1*(ALPHA / DIGIT / "-")
x-name        = "X-" [vendorid "-"] 1*(ALPHA / DIGIT / "-")
```
I can't remember if we already had to deal with those values before in the project but i'd say we have to support them. They easy and ugly way would be to accept any input, the correct one would be to check the validity of the data.

tests/event.py
C4ptainCrunch: I would say that by default, if the `._classification` attribute is `None`, the `.classification` property should return `PUBLIC` as it's the default in the RFC.
(Note, if the `._classification` attribute is `None`, it should not be added to the properties of the VEVENT when serializing [aka here](https://github.com/C4ptainCrunch/ics.py/pull/178/files#diff-0831b8f95d6c3e878c59cb1bb8cdbdd3R655))

C4ptainCrunch:
Hi @PascalBru 

Thanks for your PR, it's nice to see the project being used and worked on, even if i don't have much time to work on it :)

I just added some minor comments, if these are addressed, i'll merge the PR (and try to do a release soon after)

ics/event.py
PascalBru: Does any one have valid examples for _x-name_ and _iana-token_? This would make the implementation much more easier.

tests/event.py
PascalBru: @C4ptainCrunch You mean when reading (for example from file) or creating an event the `._classification` attribute should always be set to `PUBLIC`?

C4ptainCrunch: What i'm saying is that `._classification` (the private attribute) should be `PUBLIC` only if explicitly setted either by the user or in the source file.
When creating a new event or when reading an event that does not have the `CLASSIFICATION` property, `._classification` should be `None`

ics/event.py
C4ptainCrunch: This article can help you reading the syntax of those expressions https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form

tests/event.py
PascalBru: more test added with 0b1724b6ad772654fc78ec11c3e0f95f6a9b1ac8 
do we need to change more to resolve that comment?

PR #103, opened 2017-08-22
Add an event component RRULE

timfaner:

PR #180, opened 2019-06-18
add RRULE test cases and doc

lengthmin:
close: #103


lengthmin:
http://jakubroztocil.github.io/rrule/

this site is a good idea.

N-Coder:
Hey! We also have a second PR that adds RRULEs, #168. The difference between both is that #168 interprets what the RRULE means and adopts the processing of the event accordingly, while this PR just adds a `str` field without touching the value. Without the new field, the RRULE value would still be accessible in the `extras` Container of the Event. The only further added functionality is the automatic conversion of appropriate dicts to RRULE strings in the setter. With the upcoming version 0.8 (see #222), I hope to provide full-fledged support of RRULEs, similar to what #168 is doing. So I guess there's no big need anymore for merging this PR as an intermediate step, as there are probably no further versions in between the current 0.7 and the upcoming 0.8. Anyways, thanks for your help to both of you!

PR #189, opened 2019-08-04
Update example in README

zuphilip:
Otherwise I see the error
```
arrow.parser.ParserError: Could not match input to any of ['YYYY-MM-DD HH:mm:ss'] on '20140101 00:00:00'
```


C4ptainCrunch:
Thank you !

PR #214, opened 2019-12-27
Bugfix in Event constructor self.organizer is None

Ousret:
Hi, 👋

Thank for this great project.

With this PR I am proposing to fix : 

  - Even when passing an Organizer to Event object __init__, it is not set and therefor not serialized.

What can cause a missing Organizer in ics ?

  - For ex. it can make Outlook believe that there is not from field in a email. (Weird, isn't it ?)

Thank.

### DIFF
--- a/ics/event.py
+++ b/ics/event.py
@@ -91,7 +91,7 @@ class Event(Component):
         self._status: Optional[str] = None
         self._classification: Optional[str] = None
 
-        self.organizer: Optional[str] = None
+        self.organizer: Optional[Organizer] = organizer
         self.uid: str = uid_gen() if not uid else uid
         self.description: Optional[str] = description
         self.created: Optional[ArrowLike] = get_arrow(created)


gabelimon:
This looks right to me and it would resolve the ticket referenced above
https://github.com/C4ptainCrunch/ics.py/issues/217

N-Coder:
I'm [thinking about](https://github.com/C4ptainCrunch/ics.py/pull/222#issuecomment-582762966) replacing all the handmade dataclass initialization logic with `@attr` while I'm already at rewriting the event time logic and throwing out `arrow` (see the linked PR). Using attr would also fix this (and any similar) problem. We'll see what C4ptainCrunch thinks about that.
Unfortunately, I don't (yet) have the permission to merge and do a new release. I don't know whether this PR might/should make it into a new release on its own before the changes from the other PR are through.

C4ptainCrunch:
:wave: 

I'm giving you the rights right now :tada:

C4ptainCrunch:
I'm a bit late to the party but congratz @N-Coder for the first PR you merged yourself in ics.py :+1: 
I should really thank you a lot for your valuable help you are providing :heart: 
Thank you very much ! :pray:

PR #227, opened 2020-02-15
invert the regex matching contentline values to also allow emojis

### LINKED ISSUE #206: ParseError if emojis/unicode in description/text
I try to load a shared calendar from a Nextcloud instance via requests.
One of the calendar entries has emojis in the description, ics throws an exception when it comes accross the first emoji:

The error:
```
tatsu.exceptions.FailedToken: (1:27) expecting '\r\n' :
DESCRIPTION:Hallo zusammen😊\n\nEndlich Ist es soweit....🙃😎😁\n\n
                          ^
CRLF
contentline
start

During handling of the above exception, another exception occurred:

File "/usr/lib/python3.7/site-packages/ics/grammar/parse.py", line 78, in parse
    raise ParseError()
ics.grammar.parse.ParseError
``` 

The section of the ical:
```
BEGIN:VEVENT
CREATED:20190422T210447
DTSTAMP:20190422T210447
LAST-MODIFIED:20190422T210447
UID:NTG0P2GYEMBOTMH04Z4OP
SUMMARY:Party bei <redacted>
CLASS:PUBLIC
DESCRIPTION:Hallo zusammen😊\n\nEndlich Ist es soweit....🙃😎😁\n\n
STATUS:CONFIRMED
DTSTART;TZID=Europe/Berlin:20190601T140000
DTEND;TZID=Europe/Berlin:20190601T235500
END:VEVENT
```

### LINKED ISSUE #211: New grammar approach doesn't allow for some Unicode chars
I upgraded to ics v0.6 (still Python 3.7.1 Anaconda) after posting my last issue and found some new problems.  The EBNF file doesn't allow for Unicode characters above `\uFFFF`, such as most smileys.  (Grinning face is \u0001F601, for example.)

I've modified the `.ebnf` file with new regular expressions and that got past the smiley issue, but it stopped on an EN_DASH and I don't know why...

Here's the file content now with the old/original values commented out (I'm not saying these are correct as I haven't checked the appropriate RFCs):
```
# QSAFE_CHAR       = ?"[ \x21\x23-\x7E\u0080-\uffff]";
# QSAFE_CHAR_STAR  = ?"[ \x21\x23-\x7E\u0080-\uffff]*";
QSAFE_CHAR         = ?"[^\x00-\x1F\x22\x7F]";
QSAFE_CHAR_STAR    = ?"[^\x00-\x1F\x22\x7F]*";

# SAFE_CHAR        = ?"[ \x21\x23-\x2B\x2D-\x39\x3C-\x7E\u0080-\uffff]" ;
# SAFE_CHAR_STAR   = ?"[ \x21\x23-\x2B\x2D-\x39\x3C-\x7E\u0080-\uffff]*" ;
SAFE_CHAR          = ?"[^\x00-\x1F\x22\x2C\x3A\x3B\x7F]" ;
SAFE_CHAR_STAR     = ?"[^\x00-\x1F\x22\x2C\x3A\x3B\x7F]*" ;

# VALUE_CHAR       = ?"[ \x21-\x7E\u0080-\uffff]";
# VALUE_CHAR_STAR  = ?"[ \x21-\x7E\u0080-\uffff]*";
VALUE_CHAR         = ?"[^\x00-\x1F\x7F]";
VALUE_CHAR_STAR    = ?"[^\x00-\x1F\x7F]*";
```
Inverting the REs and selecting the characters I don't want vs. those that I do, not only does the list get shorter but the other Unicode characters are implicitly allowed.  Or should be, at least.

I've stepped through the code all the way into the GRAMMAR and a little bit deeper, but debugging a parser is not something I want to do in my spare time. ;)  Maybe the `tatsu` library has its own problems...?

I did notice that the tatsu parser uses Python's built in `re` module unless `regex` is available.  Given that the `re` module has a number of deficiencies, I'll try installing `regex` and have another go at it.

N-Coder:
contributed by @Azhrei:
"Inverting the REs and selecting the characters I don't want vs. those that I do, not only does the list get shorter but the other Unicode characters are implicitly allowed."

Fixes #206, fixes #211


N-Coder:
We should also add testcases to make sure we have no regressions regarding unicode / emoji / weird characters support in values. I'll merge once that is covered.

N-Coder:
Here are some links that could be useful for testing:
https://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-demo.txt
https://unicode.org/emoji/charts/emoji-style.txt
https://unicode.org/Public/emoji/latest/emoji-test.txt
http://www.madore.org/~david/misc/unitest/
https://character.construction/az
https://github.com/janlelis/unicode-emoji/blob/master/README.md
https://www.cl.cam.ac.uk/~mgk25/ucs/examples/TeX.txt
https://www.cl.cam.ac.uk/~mgk25/ucs/examples/combining-keycap.txt
https://www.cl.cam.ac.uk/~mgk25/ucs/examples/digraphs.txt
https://www.cl.cam.ac.uk/~mgk25/ucs/examples/postscript-utf-8.txt

N-Coder:
So I've added some test files with a lot of emoji data resulting in pretty long contentlines, but I don't have the slightest idea why tatsu fails to handle those in the CI tests.

Azhrei:
It appears to be because the text used for the test has embedded new lines and the CRLF rule is looking for CR-LF, so the rule never matches.

I don't understand why it's looking for CR-LF in the first case, since Python3 will always convert CR-LF into just `\n` on input...  However, if someone manages to get `\r\n` into the parser somehow, that CR will need to be ignored, so just changing the CRLF rule to `\n` probably isn't the right solution overall...

N-Coder:
I'm also wondering why it is looking for a line break at *that specific* place. It kind of looks like a bug in Tatsu. Im stuck here, sorry.

Azhrei:
Yeah, I don’t know tastu at all.  When I get a chance, I’ll try replacing CRLF with just LF and see if that works.  It’ll fail when both chars are actually there, but I expect that’ll be rare anyway (and good enough for my case!).

Ideally, the CR should be optional in front of LF, I just don’t know how to do that.  Normally, I’d do that in the tokenizer, not the parser.  Are they one and the same with tastu?  (See?  I don’t know anything about it!)

Thanks for getting it this far!  I may have time to play with it again late in the week...

N-Coder:
Oh yeah, I forgot that python [translates newlines](https://docs.python.org/3/library/functions.html#open-newline-parameter). Anyways, the newlines within the description are escaped ones, i.e. a backslash followed by an 'n' and not an actual newline.
The code that does the parsing is the following:
```python
CRLF = '\r\n'
grammar_path = Path(__file__).parent.joinpath('contentline.ebnf')
with open(grammar_path) as fd:
    GRAMMAR = tatsu.compile(fd.read())
# [...]
try:
    ast = GRAMMAR.parse(line + CRLF)
except tatsu.exceptions.FailedToken:
    raise ParseError()
```
So the only carriage return (followed by the only line feed) in the content line is actually appended by us. I'm slowly wondering whether using a whole EBNF engine just for the separation of content lines having the format `KEY;PARAM=PARAM_VALUE:VALUE`  is actually worth it...
Maybe we could reuse [the parser from icalendar](https://github.com/collective/icalendar/blob/master/src/icalendar/parser.py#L321) instead or revert back to our own parser that previously directly built on regexes? Or maybe we could try another PEG engine? Tatsu has the advantage of mostly following the EBNF specification, but I also read people reporting it to be pretty slow...

N-Coder:
Ahhhh I fixed it!! We forgot to include the tab in our regex, see a060afe. The [RFC](https://www.ietf.org/rfc/rfc2445.html#page-15) doesn't differentiate between "space" `\x20` and "horizontal tab" `\x09` as whitespace ("WSP") characters, so the latter also needs no escaping when appearing in content line values. But we accidentally excluded it with all the other control caracters. It also seems as if that wasn't handled in the code previously, before 0478f6120d1879988d48743373488ace06606cf4 we also only matched space but no tab character.

Azhrei:
> Anyways, the newlines within the description are escaped ones, i.e. a backslash followed by an 'n' and not an actual newline.

Hm, interesting.  I guess it makes sense, since the ics file will depict a multiline description that way, I just hadn't thought about it.

> Ahhhh I fixed it!!

Good catch!  I never read through the RFC so didn't think to compare it against the actual grammar being used.

I look forward to giving it a shot when I get a chance!  Thanks for being persistent! 😉

C4ptainCrunch:
>  I'm slowly wondering whether using a whole EBNF engine just for the separation of content lines having the format KEY;PARAM=PARAM_VALUE:VALUE is actually worth it...

It might indeed have been overkill 😞
The choice of Tatsu was indeed the EBNF-like grammar that was close the the RFC and it might not have been a great choice. Maybe we could schedule the change for something like ics 0.9 :)

N-Coder:
> Maybe we could schedule the change for something like ics 0.9 :)

I guess we can leave it in now and as long as it isn't causing more problems than it is solving there's no need to invest much time. Being able to re-use the grammar from the RFC is actually quite nice.

PR #229, opened 2020-02-15
use last compatible tatsu 4.x before python 3.8

### LINKED ISSUE #221: Doesn't work with Python < 3.8 since latest TatSu release

N-Coder:
fixes #221

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 python-dateutil
 arrow>=0.11,<0.15
 six>1.5
-tatsu>4.2
+tatsu>4.2,<5 ; python_version < '3.8'
+tatsu>4.2 ; python_version >= '3.8'


N-Coder:
Looks good in CI ([github-actions Test / build (3.7)](https://github.com/C4ptainCrunch/ics.py/pull/229/checks?check_run_id=447951157#step:4:16)):
```
Ignoring tatsu: markers 'python_version >= "3.8"' don't match your environment
[...]
Collecting tatsu<5,>4.2
  Downloading TatSu-4.4.0-py2.py3-none-any.whl (85 kB)
```
Unfortunately, we don't have automated testing for python 3.8 yet.

C4ptainCrunch:
> Unfortunately, we don't have automated testing for python 3.8 yet.

Now we do : #230 :)

N-Coder:
**Don't merge** (yet). See my comments [here](https://github.com/C4ptainCrunch/ics.py/issues/221#issuecomment-586619446). If there's no real reason to implement this, I'll both close the issue and this PR unmerged right before 0.7 is released.

PR #232, opened 2020-02-27
Fix Attendee and Organizer Parsing and Serialization

### LINKED ISSUE #207: Organizer output is wrong
```
$ pip list | grep ics | awk '{print $2}'
0.6
```

Minimal reproduction

```py
>>> from ics import Event
>>> e = Event()
>>> e.organizer = 'TRUCMACHINCHOUETTECHOSEBROLBIDULEMACHINBAZARTRUCMUCHE'
>>> print(e)
BEGIN:VEVENT
TRUCMACHINCHOUETTECHOSEBROLBIDULEMACHINBAZARTRUCMUCHE
UID:f1d79157-3483-4557-9e48-b5f27e7d7bea@f1d7.org
END:VEVENT
```
should output
```
BEGIN:VEVENT
ORGANIZER;TRUCMACHINCHOUETTECHOSEBROLBIDULEMACHINBAZARTRUCMUCHE
UID:f1d79157-3483-4557-9e48-b5f27e7d7bea@f1d7.org
END:VEVENT
```

### LINKED ISSUE #209: After parsing an iCal file, event.attendees is empty
The ical file came from a Calendar component in a Confluence page - I can probably clean up for public consumption if you need a test case.

Has this problem come up before? I thought it must be me doing something wrong.. at first.

### LINKED ISSUE #218: No way to add `partstat` or `role` to Attendee

N-Coder:
This should fix most of the issues with Attendees and Organizer by using proper Parsers and Serializaers. Note that this somewhat breaks compatibility as `str` objects are no longer allowed for these values. We might add automatic conversion and type-checking later with `attrs`.

Hopefully fixes #207, fixes #209, fixes #218, but still needs tests for this.


ics/event.py
```diff
@@ -287,51 +261,27 @@ def all_day(self):
         # the event may have an end, also given in 'day' precision
         return self._begin_precision == 'day'
 
-    def make_all_day(self, become_all_day=True):
```
aureooms: Is this based on an unmerged PR?

N-Coder:
Jep, this was unintentionally based on (the single revert-commit from) #231. Should no longer be an issue as that PR is merged now. :wink:

C4ptainCrunch:
https://github.com/C4ptainCrunch/ics.py/pull/232/files#diff-9ee99a66451cc2f8bc37e73337d765e8R200-R204 seems enough to prove that #207 is fixed, i'm writing tests for #209 and #218

C4ptainCrunch:
Adding a test for #209 made me discover a small bug (fixed in 430dcd69) but also a weird exception.

While parsing [tests/fixtures/groupscheduled.ics](https://github.com/C4ptainCrunch/ics.py/blob/master/tests/fixtures/groupscheduled.ics#L25) i get an exception: `ValueError: A ATTENDEE must have at least one CN`.
But when looking at the [RFC](https://tools.ietf.org/html/rfc5545#section-3.8.4.1), page 107 it seems to me that the CN is optional.

```

attendee   = "ATTENDEE" attparam ":" cal-address CRLF

attparam   = *(
          ;
          ; The following are OPTIONAL,
          ; but MUST NOT occur more than once.
          ;
          [...]
          (";" cnparam) / (";" dirparam) /
          [...]
          )
``` 

Is this on purpose ? The test uncovering the problem is in 76637ea

N-Coder:
> Is this on purpose ?

Not really. I must have guessed that it is required because we are using the e-mail address as default if no CN is given. So we should remove the `required=True` and maybe also the (imho unnecessary) redundant default if no CN is given?

C4ptainCrunch:
I just removed the `required=True`, for me it's OK to merge :)

mafrosis:
Great stuff! Will give this a test

PR #233, opened 2020-02-27
Add missing implementation for abstract EmailAlarm.action property

### LINKED ISSUE #213: TypeError: Can't instantiate abstract class EmailAlarm with abstract methods action
Hello, I am constantly getting this error when trying to use your library for the first time:

TypeError: Can't instantiate abstract class EmailAlarm with abstract methods action

My code:
```
# ics_reader.py

from ics import Calendar, Event
import requests

r = requests.get(url_containing_ics_file)


c = Calendar(r.text)

c
```

I can confirm certainly r.text is getting the data because separately I am able to output it.  Why this error?

N-Coder:
fixes #213, but probably should also have tests for this

### DIFF
--- a/ics/alarm/email.py
+++ b/ics/alarm/email.py
@@ -31,3 +31,7 @@ class EmailAlarm(BaseAlarm):
         self.subject = subject
         self.body = body
         self.recipients = recipients if recipients else []
+
+    @property
+    def action(self):
+        return "EMAIL"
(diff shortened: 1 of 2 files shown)


N-Coder:
Hey, @aureooms! If you're using this, do you maybe have an example file at hand from which you could generate a small test for this (maybe also covering the different other parts of e-mail alarms)? Just to make sure that this functionality has no regressions in the future.

C4ptainCrunch:
Just adding this bit of code was enough to flag the bug (shame on me :disappointed: ) so i guess it's enough to find regressions too
```python
class TestEmailAlarm(unittest.TestCase):
    def test_alarm(self):
        a = EmailAlarm(trigger=timedelta(minutes=15))
        self.assertEqual("EMAIL", a.action)
```

PR #234, opened 2020-02-27
Make sure version is given first

### LINKED ISSUE #55: The VERSION property should be the first property on the calendar.
for rfc-compliance

N-Coder:
fixes #55 and also includes a test

### DIFF
--- a/ics/serializers/icalendar_serializer.py
+++ b/ics/serializers/icalendar_serializer.py
@@ -6,7 +6,7 @@ class CalendarSerializer(Serializer):
     def serialize_0version(calendar, container):  # 0version will be sorted first
         container.append(ContentLine("VERSION", value="2.0"))
 
-    def serialize_1prodid(calendar, container): # 1prodid will be sorted second
+    def serialize_1prodid(calendar, container):  # 1prodid will be sorted second
         if calendar.creator:
             creator = calendar.creator
         else:
--- a/ics/serializers/icalendar_serializer.py
+++ b/ics/serializers/icalendar_serializer.py
@@ -6,7 +6,7 @@ class CalendarSerializer(Serializer):
     def serialize_0version(calendar, container):  # 0version will be sorted first
         container.append(ContentLine("VERSION", value="2.0"))
 
-    def serialize_prodid(calendar, container):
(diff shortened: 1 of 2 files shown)


N-Coder:
What do you mean with "fix the permutation"? Are there any further RFC requirements regarding the line order in ics files? The order is currently set to follow the alphabetic order of properties [here](https://github.com/C4ptainCrunch/ics.py/blob/4f910e6b00bb86b569e81f96743d9cc62aa47be3/ics/serializers/serializer.py#L12).

aureooms:
@N-Coder I just mean that we could make the order explicit rather than relying on the alphabetical ordering. Do we want calscale before prodid?

N-Coder:
Hmm, `PRODID` as the only other mandatory should probably come second. But I wouldn't spend too much time thinking about the exact order of all the other properties, which are order-independent for a reason. The easiest fix would be adding a `1` to `serialize_prodid`. As an alternative, we could drop the call to `sort` so that the properties appear in the same order as their serializers. But this could cause unforeseen consequences if somebody reorders the methods at some point. Honestly, I don't care that much, as long as the order is deterministic and stable between different runs and also python versions.

C4ptainCrunch:
I agree with @N-Coder, i wouldn't spend too much time filling on the ordering, using `serialize_0` and `serialize_1` seems good enough for now (and we can change it later without problem as it's an internal API.

> As an alternative, we could drop the call to `sort`

IIRC it's not as simple as it seems. I think i added the call to `sort()` because the output order was non-deterministic (maybe because dict are ordered only in recent Python versions ? Can't remember)

ics/serializers/icalendar_serializer.py
```diff
@@ -3,6 +3,9 @@
 
 
 class CalendarSerializer(Serializer):
+    def serialize_0version(calendar, container):  # 0version will be sorted first
+        container.append(ContentLine("VERSION", value="2.0"))
+
     def serialize_prodid(calendar, container):
```
C4ptainCrunch:
```suggestion
    def serialize_1prodid(calendar, container): # 1prodid will be sorted second
```

Otherwise, it's good for me :)

PR #236, opened 2020-02-29
[POC] Create a corpus of various ics files for more robust tests

C4ptainCrunch:
# Uncovered bugs
 * `ValueError: A VEVENT must have at most one CATEGORIES` (`ical4j/EstoniaHolidays.ics`) [RFC says MAY occur more than once](https://tools.ietf.org/html/rfc5545#page-53)

# Other
I might remove the incriminating file if we don't want to support that case

 * We fail if `PRODID` is not present. Maybe we should just print a warning.
 * We are case sensitive and fail when the open line has not the same case as the close (`ical4j/korganizer-lowercase.ics`): `BEGIN:Vcalendar` vs `END:vCalendaR`
 * We fail on subsecond precision date-time (eg `19970701T190000Z`). [RFC says second = 2DIGIT](https://tools.ietf.org/html/rfc5545#section-3.3.12)

# Unknowns
 * We fail on ical4j/1106817412.ics due to utf-8 wizardry ?
 * We fail on tabs inside a content line (`ical4j/mansour.ics`) does the RFC accept it ?


N-Coder:
Could you rebase this onto v0.7? That should fix the two "unknowns" as #227 allows any UTF-8 characters and also includes a fix for tabs.

C4ptainCrunch:
Indeed, the tabs issue is fixed :tada: 
For the utf-8, it even seems that it's the file reading that fails because of invalid utf-8 so i guess i'll just remove the file

N-Coder:
I have some code to pull all the mozilla thunderbird [holiday calendars](https://www.mozilla.org/projects/calendar/holidays.html) to generate a sufficiently big input file for #244, but these should also make a good addition here. I'll take care of that once the other stuff is through.

N-Coder:
Here are some further potential sources for ICS data:
https://github.com/public-apis/public-apis#calendar

N-Coder:
I guess this would still be a very good thing to keep track of (and to continue working on) regarding the test suite for 0.8...
But at least its linked from the To-Do list there some it should be findable. :man_shrugging:

PR #242, opened 2020-04-12
docs(AUTHORS.rst): Polish contents.

aureooms:
- Remove duplicated content.
  - Update some dead links.
  - Use "secure" URLs only.
  - Add details.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -32,7 +32,6 @@ Other contributors, listed alphabetically, are:
 * `@Timic3 <https://github.com/Timic3>`_
 * `@tomschr <https://github.com/tomschr>`_
 * `@Trii <https://github.com/Trii>`_
-* `@tomschr <https://github.com/tomschr>`_
 * `@zagnut007 <https://github.com/zagnut007>`_
 * `@zuphilip <https://github.com/zuphilip>`_
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -41,18 +41,17 @@ Many thanks for your contributions!
 There are also a few modules or functions incorporated from other
 authors and projects:
 
-* ``utils.iso_precision`` includes something like 10 lines of Arrow's, code,
-  which is written by Chris Smith and under Apache license
+* The function ``ics.utils.iso_precision`` includes something like 10 lines of
+  Arrow's, code, which is written by Chris Smith and under Apache license.
 
 
-A big part of the code was written at `UrLab <http://urlab.be>`_, an awesome
-hackerspace in `ULB <http://ulb.ac.be>`_ in Brussels, Belgium.
-A big part of the code was written at `UrLab <http://urlab.be>`_, an awesome hackerspace in `ULB <http://ulb.ac.be>`_ in Brussels, Belgium.
+A big part of the code was written at `UrLab <https://urlab.be>`_, an awesome
+hackerspace at `ULB <https://ulb.be>`_ in Brussels, Belgium.
 
 Thanks to:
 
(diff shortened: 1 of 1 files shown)


AUTHORS.rst
```diff
@@ -41,18 +40,17 @@ Many thanks for your contributions!
 There are also a few modules or functions incorporated from other
 authors and projects:
 
-* ``utils.iso_precision`` includes something like 10 lines of Arrow's, code,
-  which is written by Chris Smith and under Apache license
+* The function ``ics.utils.iso_precision`` includes something like 10 lines of
```
N-Coder, line 43: We can probably drop this paragraph entirely, as the `iso_precision` function is gone together with arrow (we now get the precision directly from reading the `VALUE` type parameter).

AUTHORS.rst
```diff
@@ -41,18 +40,17 @@ Many thanks for your contributions!
...
+A big part of the code was written at `UrLab <https://urlab.be>`_, an awesome
+hackerspace at `ULB <https://ulb.be>`_ in Brussels, Belgium.
 
 Thanks to:
 
-* `arrow <http://crsmithdev.com/arrow/>`_ which provides a nice API for dates, times and deltas,
-* `python-dateutil <http://labix.org/python-dateutil>`_ for parsing timezones from VTIMEZONE blocks,
-* `requests <http://python-requests.org/>`_ for giving me inspiration from it's beautiful and pythonic API (and its doc too),
-* `six <http://pythonhosted.org/six/>`_ to have made the python3 transition easier,
+* `arrow <https://arrow.readthedocs.io/en/latest>`_ which provides a nice API for dates, times and deltas,
```
N-Coder, line 52: Maybe also drop this line, as the API turned out to be not-so-nice for floating datetimes (i.e. without timezone, which is explicitly allowed by the spec). For this reason, `arrow` was completely removed and we're now using plain but correct `datetimes`. See #222.

AUTHORS.rst
```diff
@@ -41,18 +40,17 @@ Many thanks for your contributions!
...
+* `python-dateutil <https://dateutil.readthedocs.io/en/stable>`_ for parsing timezones from VTIMEZONE blocks,
```
N-Coder, line 53:
```suggestion
* `python-dateutil <https://dateutil.readthedocs.io/en/stable>`_ for parsing VTIMEZONEs and RRULEs,
```
RRULE support should be coming with #240 (see [here](https://github.com/C4ptainCrunch/ics.py/pull/240/files#diff-aaf03cd716cb3f30c8f8172b896c9719R49)).
Maybe we should also include Tatsu in this list? (#194)

AUTHORS.rst
aureooms: Yes!

AUTHORS.rst
aureooms: Yes!

AUTHORS.rst
aureooms: Yes!

AUTHORS.rst
```diff
@@ -32,7 +32,6 @@ Other contributors, listed alphabetically, are:
 * `@Timic3 <https://github.com/Timic3>`_
 * `@tomschr <https://github.com/tomschr>`_
 * `@Trii <https://github.com/Trii>`_
-* `@tomschr <https://github.com/tomschr>`_
```
aureooms, line 35: Nice catch!

AUTHORS.rst
C4ptainCrunch:
```suggestion
```

AUTHORS.rst
C4ptainCrunch:
```suggestion
```

PR #243, opened 2020-04-20
New build system

N-Coder:
For my current PR #240 I'm adding `hypothesis` as a test dependency to allow generation of random input strings to better test parsing and escaping logic (see e44e4b30b2d4438b978131a5bf1598727da8a391). I also changed how we load the grammar file, to make the package zip safe again (see 3cafe53f66c85ebd5f2d8769be5edb8d2c612f92). While working on how these could be integrated into our testsuite, I found out that `setup.py test` is deprecated ([1](https://github.com/pytest-dev/pytest/issues/5534) [2](https://tox.readthedocs.io/en/latest/example/basic.html#integration-with-setup-py-test-command)). The recommendation was to use `tox` to test the generated package directly using `pytest` instead of relying on `setup.py` to test the sources. Additionally, there's a designated successor to `setup.py` and `setup.cfg` (in which we probably accumulated a few outdated definitions), the build-system independent `pyproject.toml`. Additionally, there are newer tools like `pipenv`, `flit` and `poetry` to help with the whole build and publishing process. As I was not quite happy with how the whole development process of ics.py was set up, I wanted to give them a shot. Collecting some opinion around the internet, it seemed that `flit` was mostly targeted at very simple low-configuration projects and the development of `pipenv` somehow stagnated, while `poetry` seemed to be a well suited solution.

So this PR contains my attempt at migrating to `poetry`, with all ics.py sources moved according to the new recommended format.
It's mostly the config files in the root directory that changed, but I also removed the `dev` directory as it should no longer be needed. Next to all files from the `./ics/` directory remain unchanged and are simply moved to `./src/ics/`. I didn't copy the tests over yet, as I plan to rewrite most of them in my other branch.
The first of the two main configuration files is now `pyproject.toml`, where all meta-information on the project and how it can be installed (i.e. dependencies) and built are stored (without the need to actually execute the file and have some specific setuptools lying around). The second is `tox.ini`, where all testing related functionality is configured. A third file `.bumpversion.cfg` is from a small tool that helps with updating the version number in all places when doing a new release. The `poetry.lock` file optionally stores the dependency versions against which we want to develop, which is independent from the versions the library pulls in as dependency itself, where we are pretty liberal, and the versions we test against, which is always the latest releases. All library sourcecode now resides within a `src` folder, which is recommended as it prevents you from accidentally having the sources in your PATH when you want to test the final package.

The root directory now looks very clean and all those files have their specific purpose. If you want to configure how testing is done, you find all information in [`tox.ini`](https://github.com/N-Coder/ics.py-poetry/blob/master/tox.ini). If you want to run the tests (i.e. pytest, doctest, flake8, mypy and check that the docs build), simply run `tox` - you don't have to worry about which versions in which venvs are installed and whether you're directly testing against the sources or against a built package, tox handles all that for you. This not only comes in very handy when running the tests manually, but should also ensure that [CI](https://github.com/N-Coder/ics.py-poetry/blob/master/.github/workflows/pythonpackage.yml) does exactly the same. On a side note, we're now again publishing [coverage data](https://codecov.io/gh/N-Coder/ics.py-poetry).

If you just want to run the tests and don't need to fiddle around with the development version of ics in an interactive shell, that's all you need. For the fiddling part, just run [`poetry install`](https://python-poetry.org/docs/cli/#install) and you will have a turnkey development environment set up. Use `poetry shell` or `poetry run` to easily access the venv poetry set up for you. Publishing is now also very simple: just call `poetry publish --build` and the tool will take care of the rest. This made it very easy to make some releases on the [testing pypi instance](https://test.pypi.org/project/ics/#history).

The third and last tool you might want is `bumpversion`, if you are making new releases. But there is no need anymore to handle any venvs yourself or to install all ics.py dependencies globally. To summarize, if you want to hit the ground running and publish a new release on a newly set-up machine, the following should suffice:

```bash
git clone https://github.com/N-Coder/ics.py-poetry.git && cd ics.py-poetry
pip install tox poetry bumpversion --user
tox # make sure all the test run
bumpversion --verbose release # 0.8.0-dev -> 0.8.0 (release)
poetry build # build the package
tox --recreate # ensure that the version numbers are consistent
# check changelog and amend if necessary
git push && git push --tags
poetry publish # publish to pypi
bumpversion --verbose minor # 0.8.0 (release) -> 0.9.0-dev
git push && git push --tags
```

You can try that out if you want -- except for the publishing part maybe. Also note that `bumpversion` directly makes a commit with the new version if you don't pass `--no-commit` or `--dry-run`, but that's no problem as you can easily amend any changes you want to make, e.g. to the changelog.

The above information on developing, testing and publishing can now also be found in the docs (see CONTRIBUTING.rst). As these changes are partially based upon #240 but are also quite fundamental, I wanted to collect feedback first before including the changes into #240. The only other thing #240 is still lacking is more testing (only few files already have close to 100% coverage), and I'd prefer to provide that using `tox` in this new environment. So that's also some kind of cyclic dependency.

Sorry for the (now superfluous) issue I opened before. So @C4ptainCrunch (and maybe also @aureooms and @tomschr), what's your opinion on this?


CONTRIBUTING.rst
```diff
@@ -33,6 +33,70 @@ you are solving it. This might save you a lot of time if the maintainers
 are already working on it or have a specific idea on how the problem should
 be solved.
 
+Setting up the development environment
```
tomschr, line 36: I'd use title style in the header:

```suggestion
Setting up the Development Environment
```

CONTRIBUTING.rst
```diff
@@ -33,6 +33,70 @@ you are solving it. This might save you a lot of time if the maintainers
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+There are three python tools required to develop, test and release ics.py:
+`poetry <https://python-poetry.org/>`_ for managing virtualenvs plus dependencies and for building plus publishing the package,
+`tox <https://tox.readthedocs.io/>`_ for running the testsuite and building the documentation,
+and `bumpversion <https://pypi.org/project/bumpversion/>`_ to help with making a release.
+Their respective configuration files are ``pyproject.toml``, ``tox.ini`` and ``.bumpversion.cfg``.
+The ``poetry.lock`` file optionally locks the dependency versions against which we want to develop,
+which is independent from the versions the library pulls in when installed as a dependency itself (where we are pretty liberal),
+and the versions we test against (which is always the latest releases installed by tox).
```
tomschr, line 46: The sentence is a bit long. :wink: Maybe you should shorten it a bit. I'm not entirely sure what you want to express with the second sentence.

```suggestion
When Poetry is finished with the installation, it locks all dependencies and their versions
in the file :file`poetry.lock`.
This is independent from the versions the library pulls in when installed as a dependency itself (where we are pretty liberal),
and the versions we test against (which is always the latest releases installed by tox).
```

CONTRIBUTING.rst
```diff
@@ -33,6 +33,70 @@ you are solving it. This might save you a lot of time if the maintainers
...
+You can simply install the tools via pip:
```
tomschr, line 47:
```suggestion
Install the tools via pip:
```

CONTRIBUTING.rst
```diff
@@ -33,6 +33,70 @@ you are solving it. This might save you a lot of time if the maintainers
...
+
+.. code-block:: bash
+
+    $ pip install tox poetry bumpversion --user
+
+.. note::
+    If you want to develop using multiple different python versions, you might want to consider the
```
tomschr, line 54:
```suggestion
    If you want to develop using multiple different Python versions, you might want to consider the
```

CONTRIBUTING.rst
```diff
@@ -33,6 +33,70 @@ you are solving it. This might save you a lot of time if the maintainers
...
+    `poetry installer <https://python-poetry.org/docs/#installation>`_.
+
+    Poetry will automatically manage a virtualenv that you can use for developing.
+    By default, it will be located centrally in your home directory (e.g. in ``/home/user/.cache/pypoetry/virtualenvs/``).
+    To make poetry use a ``./.venv/`` directory within the ics.py folder use the following config:
```
tomschr, line 59:
```suggestion
    To make poetry use a ``./.venv/`` directory within the ics.py folder, use the following config:
```

CONTRIBUTING.rst
```diff
@@ -33,6 +33,70 @@ you are solving it. This might save you a lot of time if the maintainers
...
+
+Now you are ready to setup your development environment using the following command:
+
+.. code-block:: bash
+
+    $ poetry install
+
+This will create a new virtualenv and install the dependencies for using ics.py.
+Furthermore, the current source of the ics.py package will be available similar to running ``./setup.py develop``.
+To access the virtualenv, simply use ``poetry run python`` or ``poetry shell``.
```
tomschr, line 73:
```suggestion
To access the virtualenv, use ``poetry run python`` or ``poetry shell``.
```

CONTRIBUTING.rst
```diff
@@ -33,6 +33,70 @@ you are solving it. This might save you a lot of time if the maintainers
...
+If you made some changes and now want to lint your code, run the testsuite, or build the documentation, simply call tox.
```
tomschr, line 74:
```suggestion
If you made some changes and now want to lint your code, run the testsuite, or build the documentation, run tox.
```

CONTRIBUTING.rst
```diff
@@ -33,6 +33,70 @@ you are solving it. This might save you a lot of time if the maintainers
...
+You don't have to worry about which versions in which venvs are installed and whether you're directly testing against the sources or against a built package, tox handles all that for you:
+
+.. code-block:: bash
+
+    $ tox
+
+To just run a single task and not the whole testsuite, use the ``-e`` flag:
```
tomschr, line 81:
```suggestion
To run a single task and not the whole testsuite, use the ``-e`` flag:
```

CONTRIBUTING.rst
```diff
@@ -33,6 +33,70 @@ you are solving it. This might save you a lot of time if the maintainers
...
+
+.. code-block:: bash
+
+    $ tox -e docs
+
+Checkout ``tox.ini`` to see which tasks are available.
```
tomschr, line 87: IMHO we should avoid this advise. This can be done with tox itself. I would recommend to use `tox-av`. As such I would rewrite this paragraph as:

```suggestion
To get a list of all available tasks, run :command:`tox -av`.
```

CONTRIBUTING.rst
```diff
@@ -66,3 +130,47 @@ Last thing
 
   The title of your PR will become the commit message, please craft it
   with care.
+
+How to make a new release
+-------------------------
+
+If you want to hit the ground running and publish a new release on a freshly set-up machine, the following should suffice:
```
tomschr, line 137: I'm not sure what you mean by "ground running" and why it is important to do that on a "freshly set-up machine".
Would the following sentence be sufficient enough?

```suggestion
If you want to publish a new release, use the following steps
```

CONTRIBUTING.rst
```diff
@@ -66,3 +130,47 @@ Last thing
...
+    bumpversion --verbose minor # 0.8.0 (release) -> 0.9.0-dev
+    # Start new changelog
+    vi CHANGELOG.rst && git commit -i CHANGELOG.rst --amend
+    # Publish to GitHub
+    git push && git push --tags
+
+Please note that bumpversion directly makes a commit with the new version if you don't
+pass ``--no-commit`` or ``--dry-run``,
+but that's no problem as you can easily amend any changes you want to make.
+Further things to check:
```
tomschr, line 172: I'm not sure about the "but that's no problem". You mean probably to amend an existing commit with `git commit --amend`, right? Hmn, I would avoid that information. A commit done by bumpversion should be not diluted with something else.

```suggestion
Note, that bumpversion directly makes a commit when bumping a new version. If you
want to avoid this, add the option ``--no-commit`` or ``--dry-run`` to the command.

Further things to check:
```

tox.ini
```diff
@@ -0,0 +1,76 @@
+[tox]
+isolated_build = true
+envlist = py36, py37, py38, flake8, mypy, docs
+
+[testenv]
+extras =
```
tomschr, line 6: I would propose to add a `description` keyword to describe this target. This is very helpful if you want to get a list of all available task and their description with `tox -av` as describe in one of my comments above.

```suggestion
description = Run the test suite
extras =
```

Same with the other targets. :wink:

tox.ini
```diff
@@ -0,0 +1,76 @@
...
+commands =
+    flake8 --version
+    flake8 src/
+
+[testenv:mypy]
+basepython = python3.8
+commands =
+    mypy -V
+    mypy --config-file=tox.ini src/
+
```
tomschr, line 24: Why not combine both flake8 and mypy and call them in an additional target `checks` together? The nice thing about this is, you can always amend this new `checks` target with additional tasks. For example, `docformatter` or checking something else. It's also easier to document it

In one of my `tox.ini` files I have this structure (adapted to your existing targets):

```suggestion

[testenv:checks]
description = Run code style checks
basepython = python3
deps =
    {[testenv:flake8]deps}
    {[testenv:mypy]deps}
commands =
    {[testenv:flake8]commands}
    {[testenv:mypy]commands}

```

When you run `tox -e checks` you run both `flake8` and `mypy`. Voila! :wink:

.gitignore
```diff
@@ -1,45 +1,113 @@
-# Compiled python modules.
```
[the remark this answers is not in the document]
N-Coder: eff8add6fc5d7dfc875473c029e7f8e5d74f5c19 now uses the github version

CONTRIBUTING.rst
N-Coder: The file locks the versions of dependencies in the development environment set up by poetry, so that this environment is the same for everyone. The file is only read by poetry and not included in any distributions, thus these restrictions don't apply when running `pip install ics`. As tox manages its own environments and also doesn't read the lock file, it just installs the latest versions of dependencies for testing. More details on the poetry side can be found here:
https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control

We don't necessarily have to include the file in the repo (see the link above), and I also have no strong opinion here#, I just thought that it could be nice to have for developing without any side-effects for distributions and testing.

pyproject.toml
```diff
@@ -0,0 +1,52 @@
...
+# [tool.poetry.dev-dependencies]
+pytest =                     { version = "^5.2",    optional = true }
+pytest-cov =                 { version = "^2.8.1",  optional = true }
+hypothesis =                 { version = "^5.8.0",  optional = true }
+mypy =                       { version = ">=0.770", optional = true }
+flake8 =                     { version = "^3.7.9",  optional = true }
+sphinx =                     { version = "^3.0.0",  optional = true }
+sphinxcontrib-napoleon =     { version = "^0.7",    optional = true }
+sphinx-autodoc-typehints =   { version = "^1.10.3", optional = true }
+
```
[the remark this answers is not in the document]
N-Coder: `tox` is also missing from this listing, which is somewhat due to the chicken-and-egg problem of specifying requirements regarding the environment used for building/installation (in constrast to the runtime environment, for which requirements can easily specified the usual way). I thought its probably best to have `poetry` and `tox` in the system/user-wide python installation, as these tools are directly called by us and are also able to manage their version on their own (i.e. they seem to automatically switch to a newer version themselves if required in their respective config files). Additionally having a poetry venv with a tox installation, which is then used to create a tox venv in which the poetry-installed tox installs ics.py for testing seemed like unnecessary nesting, especially as tox again installs poetry in its managed venv to build our package.

The case isn't as clear for bumpversion. I thought it fitted nicely into the `pip install tox poetry bumpversion --user` required to set up the tools used for publishing. We also have no big requirements regarding its version as the current, pretty old one works perfectly fine (so I also see no big reason to switch to bump2version now and then back once they make it to merge). Basically the difference is between having it in the global env managed by the user or in the poetry managed venv and between running `bumpversion` and `poetry run bumpversion`. I slightly preferred the former, as `tox` and `poetry` are used in the same way.

CONTRIBUTING.rst
N-Coder: Yeah, `git commit --amend` was what I was thinking about, as I thought all release-related information being collected in a single commit would be nice. If that's not what we want, we should probably also change the `vi CHANGELOG.rst && git commit -i CHANGELOG.rst --amend` bash code / steps above.

tox.ini
N-Coder: Done :+1:

tox.ini
N-Coder: That's cool, thanks! Once we got our docstrings to a presentable state, we might even consider having their style checked by a stricter tool like [pydocstyle](http://www.pydocstyle.org/en/stable/), which also checks the docstring contents.

N-Coder:
> I've focused more on infra and documentation files as I can't tell much about the source code. Probably I've left enough comments for you to digest. Hope you find them useful.

@tomschr thanks a lot for checking that! This PR is all about infra and having a nicely coded project won't help anyone else if the documentation is no use. :)

Regarding the actual source code changes in this PR, most of them are there because I made these changes against my local version and not a fresh clone, so that's mostly small stuff that doesn't quite belong here, but which I would have to commited to the other branch anyways. They're about checker happyness and contain no functional changes, so it's probably not worth sieving them out manually, as the commits will get squashed in the end.

N-Coder:
@C4ptainCrunch, @aureooms do you have an opinion on this? This is currently somewhat blocking my work on 0.8 and it would be nice if I could merge this and continue working on the parser next weekend or so.

aureooms:
> @C4ptainCrunch, @aureooms do you have an opinion on this? This is currently somewhat blocking my work on 0.8 and it would be nice if I could merge this and continue working on the parser next weekend or so.

@N-Coder Looks fines to me. Great work!

CONTRIBUTING.rst
```diff
@@ -89,15 +91,16 @@ To run a single task and not the whole testsuite, use the ``-e`` flag:
 To get a list of all available tasks, run :command:`tox -av`.
 
 .. note::
-    If you want to run any tasks of tox manually, you need to make sure that you also have all the dependencies of the task installed.
-    This is easily ensured by also installing the "dev" extra dependencies into you main environment:
+    If you want to run any tasks of tox manually, you need to make sure that you also have all the testing dependencies of the task listed in ``tox.ini`` installed.
```
tomschr, line 94: I think, "listed" and "installed" doesn't really match.

```suggestion
    If you want to run any tasks of tox manually, make sure you have all the testing dependencies of the task listed in :file:`tox.ini`.
```

CONTRIBUTING.rst
```diff
@@ -73,6 +70,11 @@ Now you are ready to setup your development environment using the following comm
 This will create a new virtualenv and install the dependencies for using ics.py.
 Furthermore, the current source of the ics.py package will be available similar to running ``./setup.py develop``.
 To access the virtualenv, use ``poetry run python`` or ``poetry shell``.
+The :file:`poetry.lock` file locks the versions of dependencies in the development environment set up by poetry, so that this environment is the same for everyone.
+The file is only read by poetry and not included in any distributions, so these restrictions don't apply when running ``pip install ics``.
+As tox manages its own environments and also doesn't read the lock file, it just installs the latest versions of dependencies for testing.
+More details on the poetry side can be found `here <https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control>`_.
+
```
tomschr, line 77: I'd have a small suggestion. :wink: I would shorten some sentences, remove weasel words etc. so here is the slightly changed version:

```suggestion
The :file:`poetry.lock` file locks the versions of dependencies in the development environment set up by poetry. This ensures that such an environment is the same for everyone.
The file :file:`poetry.lock` is only read by poetry and not included in any distributions. These restrictions don't apply when running ``pip install ics``.
As tox manages its own environments and doesn't read the lock file, it installs the latest versions of dependencies for testing.
More details on the poetry side can be found in the `poetry documentation <https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control>`_.

```

pyproject.toml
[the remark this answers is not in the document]
N-Coder: They are now included as extra. If you really want to have tox (and bump2version) managed by poetry, you can grab them via `poetry install -E dev`, but I'd recommend to have the installed on the same level as poetry.

.bumpversion.cfg
```diff
@@ -0,0 +1,23 @@
+[bumpversion]
```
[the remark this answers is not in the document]
N-Coder: done, we're directly using bump2version now as the bumpversion now only pulls in bump2version as dependency

Issue #244, opened 2020-05-10
V0.7 is much slower to parse large calendar feed then V0.5


cfhowes:
I have the following code:

```
    resp = requests.get(ical_feed_url)
    if resp.status_code != 200:
        logger.error('> Error retrieving iCal feed!')
        return None

    try:
        print(f'begin parse {datetime.datetime.now()}')
        cal = ics.Calendar(resp.text)
        print(f'end parse {datetime.datetime.now()}')
    except Exception as e:
        logger.error('> Error parsing iCal data ({})'.format(e))
        return None
```
I recently tried upgrading from 0.5 to 0.7 but found that it was much slower, and eats a ton more CPU.  See my print logs from a few minutes apart on the same feed, same computer.

# V0.7 timing:
```
begin parse 2020-05-09 18:08:26.872346
end parse 2020-05-09 18:10:20.411268
```

# V0.5 timing:
```
begin parse 2020-05-09 18:12:34.763552
end parse 2020-05-09 18:12:38.981032
```

I am using python 3.6.8, Ubuntu 16.04.

Due to privacy concerns I'm not going to include the ical feed, but it has about 15,000 events in it. I am happy to run some tests or try things out if i get pointed in the right direction.

aureooms:
Can you check if the same happens with v0.6 ? That is where tatsu parsing was introduced.

N-Coder:
Could you check how long [calendar_string_to_containers](https://github.com/C4ptainCrunch/ics.py/blob/f6cb12281bf78e210d51c6ff5bd324d8f2c01aef/ics/grammar/parse.py#L209) takes when you pass it the `resp.text` ICS string, compared to calling `ics.Callendar`? The method `calendar_string_to_containers` calls Tatsu to parse every line of the string into a `ContentLine`. The `ics.Calendar` constructor first calls this, but then proceeds to read/parse the `ContentLine`s to python objects and stores them in the fields of the calendar. This second step still has some (known) minor issues in 0.7, which might result in increased runtime or memory consumption for big inputs. Tatsu in the past also caused some problems with taking to much time, so comparing the two durations should help decide which of the two steps causes the issues. Could you also tell us how many bytes and lines the file has?

cfhowes:
Thanks for the info, here's my results:

cal file:
```
ls -la maycal.ics
-rw-rw-r-- 1 cfhowes cfhowes 5551179 May 10 13:28 maycal.ics
more maycal.ics | wc -l
173189
```

[...]

PR #247, opened 2020-06-01
Contentline parser

N-Coder:
In addition to optimizing the Tatsu Grammar and precompiling it, I included two further alternative parsers in this branch: one that is based on handwritten character searching (using the fast `str.index` and `str.find` functions without unnecessarily slicing substrings) and one that is based on the alternative [regex module](https://pypi.org/project/regex/) and its full history of group captures in `regex.Match.captures` (while `re.Match.group` only stores the latest capturing group). Some first tests indicate that both similarly outperfom tatsu, with the optimized tatsu taking roughly a minute with the reference file I generated for #244 and both alternative parsers taking roughly a second. The profiler also indicated that this is close to the possible optimum, as Python object initialization and access consumed the biggest part of that one second. I'll run some more (performance) tests and also test with malformed input and then report back to decide how to use these parsers in v0.8.

Note: the `regex` parser currently doesn't work on pypy3 due to this unicode [bug](https://bitbucket.org/mrabarnett/mrab-regex/issues/374/split-offset-wrong-due-to-utf-8-issues-on) somewhere in those two components.


N-Coder:
I ran some performance test with 
- three different parsers: handwritten, tatsu, and regex-based
- reading the input string from memory or harddisk
- line-wise iterating over the input line-wise or reading it as one big string
- using the built-in `re` module (only for handwritten and tatsu) or using the extended `regex` module
- using CPython or PyPy
- using a precompiled tatsu grammar or loading it from the ebnf file during runtime.

The results can be summarized as follows:
- Tatsu is by far slower than the others. The fastest run was line-wise reading using a precompiled grammar on PyPy, which took about 7 seconds. Switching to CPython or reading the file as a whole each add *an order of magnitude* to that (yep, the fastest tatsu run on CPython was 49s), while not precompiling only adds a minor overhead. The other parsers all took less than 5s no matter what interpreter was used. 
- Using the built-in `re` module is a little bit faster than the extended `regex` module, while also not failing with weird unicode symbols (see the `regex` bug mentioned above). Similarly, the handwritten parser is a little bit faster than the one relying on the extended `regex` functionality.
- Reading line-wise instead of the whole file at once is marginally faster and needs a little less memory (unsurprisingly), but the numbers vary a little bit depending on whether we read from memory or from disk.

See b85e0a4 for the code used for testing. Note that I didn't test much with invalid data, but I guess the results for the good case are clear enough (and tatsu is even more prone to spending too much time on broken inputs).

Supporting three different parser implementations and giving the user the possibility to choose from them adds a lot of complexity on both sides. As the handwritten parser is faster than the other two, I decided to only ship that one to the users, drastically simplifying the code in cd97f4b. The `regex` parser is no longer needed and the tatsu parser has been moved to the testing code, so that we can use it as reference when parsing files from the corpus.

aureooms (approved):

Issue #245, opened 2020-05-16
Roadmap for v0.8


N-Coder:
This is a meta-issue to collect all the things we want to have in v0.8 (see also the [milestone](https://github.com/C4ptainCrunch/ics.py/milestone/8)) and how we want to get there. I'll try to keep this list updated when things change over time.

# Major Changes
- [x] remove arrow, transition to attrs and add Timespan (PR #222, including the [old ToDo list](https://github.com/C4ptainCrunch/ics.py/pull/222#issuecomment-583764888))
- [ ] migration to new data conversion implementation (PR #240)
- [ ] new build system (PR #243)
- [ ] improved documentation (#220, #225, PR #223)
- [ ] proper timezone handling (PR tbd, #129, #161)
- [ ] faster ContentLine parser implementation (PR tbd, #244)

# Open ToDos, Issues and Bugs
- [ ] finally working all_day event support (#155, #173, #92, )
- [ ] real world tests (#236)
- [ ] migration of old test suite
- [ ] full test coverage for new code
- [ ] tests preventing regressions on old bugs
  #181, #182, #188, ...

# Wishlist
- [ ] recurring events (#14, #168)
- [ ] parsing / serialization hooks and customizations
- [ ] varying levels of strictness (validate vs. try hard to interpolate the right data)
- [ ] line folding (#215, #210)
- [ ] webcal
- [ ] parity with all RFC 5545 properties (#115)

waynew:
Has this been abandoned or is there still progress happening to this release?

N-Coder:
It's definitely not abandoned, I'm still hoping to make this release happen.
The biggest blocker is probably the timezone handling, which proved to be [more complicated](https://github.com/ics-py/ics-py/pull/249#issuecomment-774755657) than I initially thought.
I guess my preferred approach would now be to have as little complications in that area as possible (i.e. not serialize anything if not explicitly given all the data) and maybe deliver the missing parts in a later release.
Of course only speaking to myself in that thread and not getting any feedback (or even help) from anyone else isn't exactly motivating.
Other than that, my main motivation for working on this was to get all my private meetings and events properly managed, but that issue was pretty much solved in another, less enjoyable way a little over a year ago, while work also got a lot more stressfull.
Maybe once life comes back to normal in a hopefully not too distant future, I (and hopefully also others - maybe you?) will again have more time and motivation to work on this.

make-github-pseudonymous-again:
[...]

PR #249, opened 2020-06-13
improved ComponentMeta system and Timezone conversion

### LINKED ISSUE #129: Add VTIMEZONE / tzinfo Serialization and better TZID Parsing
Do we provide VTIMEZONE support while creating ics file ?
If yes then kindly refer which API needs to get called ?

N-Coder:
This PR contains two big things:
1. a revamped ComponentMeta / Converter system that uses less magic than before, but should be easier to extend and set-up reliably (see f333e1a, [`initialize_converters`](https://github.com/C4ptainCrunch/ics.py/commit/5dd25b56908bc3633af7f0469840731e07ced05b#diff-43120e91d8c9de5c704707963744ac27R13); and the customizations of [Alarm](https://github.com/C4ptainCrunch/ics.py/commit/5dd25b56908bc3633af7f0469840731e07ced05b#diff-bc83fdecc188dd68c5724c812689b9c6R76) and [Calendar](https://github.com/C4ptainCrunch/ics.py/commit/f333e1a223dd79141785e2915a01c83dbe8e5d05#diff-ccdc9b8c0fa3ae08149d9f2b9ba8b214R32) for examples)
2. some work towards reliable parsing, interpretation and serialization of Timezone IDs (i.e. Olson and Windows) and tzinfo objects (i.e. builtin, dateutil and pytz) in ca00173. This should fix #129, but there's also still quite a lot of work to do, mostly handling all the special cases correctly, working on how to include the required data files and ensuring that everything does what it should.

I also started to include tests against regressions of prior github issues (see e8e13e4, where mostly Timezone-related issues are covered) and some more general polish.


N-Coder:
@aureooms [asked](https://github.com/C4ptainCrunch/ics.py/issues/265#issuecomment-716486762):
> Can you explain precisely what kind of layer would ics_vtimezones add between ics and these data files. What are the sources of these files? Everything comes from Olson timezone database? Why not call it olson instead of ics_vtimezones?
> I can see a few python packages that seem to bundle these files (https://pypi.org/search/?q=olson). None of these are usable?

I explained my initial thoughts [here](https://github.com/C4ptainCrunch/ics.py/issues/129#issuecomment-609069642). The [README of the ics_vtimezones project](https://github.com/N-Coder/ics_vtimezones/) also has some relevant info on what's packaged there.

Basically, to get really portable standards-compliant ics files, you need to include the full vTimezone (i.e. usually the recurring changes of UTC-offset when changing between daylight-saving and normal time for that region, but also any political decisions changing the offset at specific points in time) specification of any timezone used within the file. There are two common sources for this information: the IANA time zone database (also informally known as [Olson database](https://en.wikipedia.org/wiki/Tz_database)) or the time zone database [shipped by windows](https://docs.microsoft.com/en-ca/archive/blogs/bclteam/exploring-windows-time-zones-with-system-timezoneinfo-josh-free). Please note that both databases change over time - e.g. there have been [4 updated versions](https://github.com/timparenti/tzdata-meta/blob/master/_data/years/2020.yml) of the IANA time zone database released in 2020 alone. The IANA time zone database is shipped on most unix systems at different paths and also used in many programming languages (e.g. the `pytz` and the new stdlib 3.9 [`zoneinfo`](https://docs.python.org/3/library/zoneinfo.html) and [`tzdata`](https://pypi.org/project/tzdata/) modules for python, see also the list [here](https://data.iana.org/time-zones/tz-link.html#compilers)), but is not always available on windows systems. While there is a mapping between windows and IANA timezone names, most ecosystems [ship](https://en.wikipedia.org/wiki/Tz_database#Use_in_software_systems) their own portable package with the IANA database:

> The Olson timezone IDs are also used by the Unicode Common Locale Data Repository (CLDR) and International Components for Unicode (ICU). For example, the CLDR Windows–Tzid table maps Microsoft Windows time zone IDs to the standard Olson names, although such a mapping cannot be perfect because the number of time zones in Windows systems is significantly lower that those in the IANA TZ database.

So you are right, most UNIX systems already have (some revision of) that database installed, there are at least two python modules that package this information and starting from python 3.9, the database will also be used by the standard library.

The issue is that just having the database files available doesn't help us, as those files are usually in a specific binary format, which you first need to parse and then convert into an ics-compatible vTimezone representation. IANA has a long [list](https://data.iana.org/time-zones/tz-link.html#compilers) of tools that can be useful for this. Here's a list of stuff that might be useful for us:
- A python parser implementation is only since recently widely available with the backported python 3.9 [`zoneinfo` module](https://github.com/pganssle/zoneinfo/blob/master/src/backports/zoneinfo/_zoneinfo.py).
- `pytz` ships the timezone data [precompiled into python objects](https://github.com/newvem/pytz/blob/master/pytz/zoneinfo/CET.py) and I couldn't find what they use for parsing and precompiling.
- The other python ics library `vobject` seems to [use PyICU](https://github.com/eventable/vobject/blob/master/vobject/change_tz.py), which is "a Python extension wrapping the ICU C++ libraries", and then [probe](https://github.com/eventable/vobject/blob/master/vobject/icalendar.py#L93) any given `tzinfo` objects for transitions.
- Apple's python calendar server has its own python [converter implementation](https://github.com/apple/ccs-pycalendar/blob/master/src/zonal/tzconvert.py).
- The C library `libical` has its own C program for parsing and generating the ics vTimezone definitions it ships, [`vzic`](https://github.com/libical/vzic).
- The Java `ical4j` library does a [similar](https://github.com/ical4j/ical4j#ical4j-and-timezones) [thing](https://github.com/benfortuna/tzurl) (I also listed both in my initial thoughts posting).
- Babel includes the CLDR timezone mapping, but unfortunately only allows lookups in the [wrong direction](http://babel.pocoo.org/en/latest/api/dates.html#babel.dates.get_timezone_name).

The issue with `zoneinfo` and `pytz` is, similar to any other `tzinfo` implementation, that the underlying information is not directly available (in the case of the C implementation of `zoneinfo`, this information is actually not available from python at all). In case of `pytz` and the pure-python version of `zoneinfo`, we could still try to grab this internal data and hope that its representation never changes, breaking our somewhat hacky code. As an alternative, we could also ship our own parser to be sure that the interface never changes. In both cases, we could then reuse the logic to convert the data to ics vTimezones at runtime from other projects. Of course, both approaches would need a lot more code on our side than just using the ready-made `vzic` tool to pregenerate the data and ship it as its own package. On the other hand, this package also need regular maintenance whenever there are new releases of the IANA database. I went with the packaged approach before the python maintainers decided to go with a standard `zoneinfo` and `tzdata` module, so I might give generating this data at runtime another try and see how much additional (and how hacky) code on our side that needs.

N-Coder:
I created an overview how different other tools handle the inclusion of vTimezone definitions when writing ics files:
https://github.com/N-Coder/vtimezone-examples

There seem to be three different broad categories
1. don't include vTimezone definition at all:
us currently, `python-icalendar`, google calendar
2. include some specific transition dates (as available in the compiled files in `/usr/share/zoneinfo/`):
`python-icalendar-khal`, `python-vobject`, outlook (with the Windows TZ names)
3. include the actual transition rrules of the timezone (which are usually only available from the sources of the olson database, which doesn't ship with Unix distros or the `zoneinfo` module - see `tzdb.txt` vs `zoneinfo.txt`):
`apple-ccs-pycalendar-zonal`, `tzurl(-outlook)`, `vzic`, apple calendar, evolution (Ximian)

Using some minor introspection and the publicly available API, I was able to generate all the data for case 2. from `pytz` and `zoneinfo` instances (see `zoneinfo.txt` and `convert_zoneinfo.py`), which pretty much covers the years 1880 up to 2040, but gives all the transition dates explicitly, as the compiled timezone files don't contain meta information about the rules behind the transition dates. As the two versions of the tzurl files show, all those different transition dates might confuse Outlook. To get to the cleaner and shorter rrule definitions serialized in case 3, you need to have the Olson tzdb source files at hand - which is what I did to create the data for the `ics_vtimezones` project. As an alternative, these ics files are also served by the tzurl project via HTTP and could be downloaded on the fly, offloading the work to keep them up to date to someone else. In both cases, `vzic` is used to convert the tzdb rule files to ics vTimezones.

To summarize, none of the three different behaviours seem perfect. Maybe we should select a sensible default for ics.py and provide opt-in configurations for the alternatives:
1. don't include vTimezone definitions
2. include vTimezone defintions obtained by introspection from the `pytz` / `zoneinfo` tzinfo instances, i.e. the compiled timezone files only containing specific instances
3. include proper `rrule`-based vTimezone definitions...
    a. provided by the `ics_vtimezones` package, if installed 
    b. download `tzurl` vTimezone files on the fly

Note this only applies to writing ics files, interpreting and also serializing vTimezone definitions when they were already included in a parsed ics file is pretty straight-forward.

waynew:
I think a graceful upgrade is a good approach - IIUC, the vtimezone needs to specify RRULE and whatnot?

Though, if you can specify e.g. 'America/Chicago' then that should be the default.

If not, dumping as UTC with a warning about losing daylight info would be appropriate, and gracefully upgrading to use a library that can add the vtimezone info

N-Coder:
Hmm, that was closer to a usable state than I thought. I kept the code using `ics_vtimezones` to provide the vTimezone definitions in, as I figured that making it optional and allowing to switch these providers would make things a lot more complicated. So we should be good for now with depending on `ics_vtimezones`, but we can think about allowing alternatives later. Also, the current version on PyPI is already slightly outdated, but it should still be good for almost all use-cases. I guess the built-in `zoneinfo` module doesn't get updates every few months either. Still, we could reduce the maintenance overhead for  `ics_vtimezones` and ensure that it actually stays up to date by using a weekly-scheduled GitHub CI action that checks for new data and automatically publishes a new version if there is any, see [ics_vtimezones#2](https://github.com/N-Coder/ics_vtimezones/issues/2).

The code could still use some more testing and the component system would largely benefit from a self-contained documentation, but the core of this PR should be good to go. So I'd merge this soonish so that other work can easily build on this, meaning that we can also switch to working on the remaining points in #245 (i.e. mostly testing and documentation) more step-by-step in more fine grained PRs. @C4ptainCrunch due to the size of this PR I'm against again squashing into one uber-commit, I'd prefer fast-forwarding master to contain all the commits separately, but unfortunately I can't do that due to missing permissions. Could you make me co-owner of the organisation/repo so that I can merge this PR properly and also continue working on things like your PR #274?

make-github-pseudonymous-again:
I agree with not squashing.

plammens:
When serializing a timezone, the following is raising an `AttributeError` since the `tzinfo` object doesn't have a `_name` attribute:

https://github.com/ics-py/ics-py/blob/b551ca566f0f10256b1423f7a2d774a473de826f/src/ics/timezone/converters.py#L93

Minimal example:

```
import datetime
import ics

# make a datetime with an "anonymous" timezone
dt = datetime.datetime.fromisoformat("2017-10-01T12:00:00+09:00")
print(dt.tzinfo)  # datetime.timezone(datetime.timedelta(seconds=32400))
print(hasattr(dt.tzinfo, "_name"))  # False
print(hasattr(dt.tzinfo, "_offset"))  # False

event = ics.Event(begin=dt, duration=datetime.timedelta(minutes=30))
event.serialize()  # AttributeError
```

N-Coder:
Unfortunately, `tzname` returns the name for *that specific datetime*, which doesn't actually indicate the desired timezone including any transitions:
```python
>>> from dateutil.tz import gettz
>>> from datetime import datetime, timedelta
>>> now = datetime.now(tz=gettz("Europe/Berlin"))
>>> later = now + timedelta(days=183)
>>> now.tzname(), now.utcoffset(), now.tzinfo
('CEST', datetime.timedelta(seconds=7200), tzfile('/usr/share/zoneinfo/Europe/Berlin'))
>>> later.tzname(), later.utcoffset(), later.tzinfo
('CET', datetime.timedelta(seconds=3600), tzfile('/usr/share/zoneinfo/Europe/Berlin'))
```
So for more complex timezone objects, we get 'CEST' or 'CET' depending on the timestamp in this case, while the user would want and expect 'Europe/Berlin' - working with 'CEST' or 'CET' might even yield wrong results once you start modifying datetimes based on that fixed timezone. So to get the information we actually want, we unfortunately need to touch some internals as there is no suitable API to get this information. See [here](https://github.com/ics-py/ics-py/issues/129#issuecomment-609069642) and [here](https://github.com/ics-py/ics-py/pull/249#issuecomment-719942261) for more background information on why there seems to be no other way how to do this.

It seems that in your specific case (maybe starting with Python 3.9), the class was replaced with a built-in / compiled C++ implementation that replaces the python implementation for efficiency reasons:
```python
>>> dt = datetime.fromisoformat("2017-10-01T12:00:00+09:00")
>>> dt.tzname(), dt.utcoffset(), dt.tzinfo
('UTC+09:00', datetime.timedelta(seconds=32400), datetime.timezone(datetime.timedelta(seconds=32400)))
>>> print(inspect.getsource(type(dt.tzinfo)))
class timezone(tzinfo):
    __slots__ = '_offset', '_name'
    ...
>>> dt.tzinfo.utcoffset
<built-in method utcoffset of datetime.timezone object at 0x7f917c2a9750>
```
Fortunately, in this case we can use the class methods of the `datetime.timezone` type instead of direct access to the properties to get suitable information.
```python
>>> dt.tzinfo.tzname(None)
'UTC+09:00'
>>> dt.tzinfo.utcoffset(None)
datetime.timedelta(seconds=32400)
```
I'll push a small fix for this, could you then try again?

---

Thanks for the hint with `@functools.lru_cache`! The information which timezones are used in one `Calendar` and how they are defined is very local to the respective `Calendar` instance, so reusing them between calls might lead to various problems (not only regarding multi-threading, but also name-clashes, different versions of the same timezone, ...), so having a (de-)serialization-local `context` seems to be the way to go.

PR #255, opened 2020-08-23
Fix some code quality and bug-risk issues

pnijhara:
# Summary of changes

- Use literal syntax instead of function calls to create the data structure
- Move statements from else clause to the loop body
- Remove bad comparison test

Find other issues here → https://deepsource.io/gh/pnijhara/ics.py/issues/

---

Also added a `.deepsource.toml` configuration file to run DeepSource analysis on the repo with. Upon enabling DeepSource, the analysis will run on every PR and commit to detect various types of code quality issues in the changes.

To enable [DeepSource](https://deepsource.io/)  analysis after merging this PR, please follow these steps:

1. [Signup](https://deepsource.io/signup/) on DeepSource with your GitHub account and grant access to this repo.
2. Activate analysis on this repo [here](https://deepsource.io/gh/C4ptainCrunch/ics.py).

You can also look at the docs for more details. Do let me know if I can be of any help!

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -8,7 +8,7 @@ import ics
 def test_version_matches():
     dist = pkg_resources.get_distribution('ics')
     print(repr(dist), dist.__dict__, sys.path, ics.__path__)
-    assert len(ics.__path__) == 1
+    assert len(ics.__path__) is 1
     ics_path = ics.__path__[0]
     assert "/site-packages/" in ics_path and not "/src" in ics_path, \
         "ics should be imported from package not from sources '%s' for testing" % ics_path
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,11 @@
+version = 1
+
+test_patterns = ["tests/**"]
+
+[[analyzers]]
+name = "python"
+enabled = true
+
+  [analyzers.meta]
+  runtime_version = "3.x.x"
+  
--- a/src/ics/converter/value.py
+++ b/src/ics/converter/value.py
@@ -88,8 +88,8 @@ class AttributeValueConverter(AttributeConverter):
             if nr > 0:
                 params["VALUE"] = [converter.ics_type]
             return converter
-        else:
-            raise ValueError("can't convert %s with %s" % (value, self))
+            
+        raise ValueError("can't convert %s with %s" % (value, self))
 
     def serialize(self, component: "Component", output: Container, context: ContextDict):
         if self.is_multi_value:
--- a/src/ics/icalendar.py
+++ b/src/ics/icalendar.py
@@ -57,9 +57,9 @@ class Calendar(CalendarAttrs):
             creator (**string**): uid of the creator program.
         """
         if events is None:
-            events = tuple()
+            events = ()
         if todos is None:
-            todos = tuple()
+            todos = ()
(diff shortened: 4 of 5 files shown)


tests/__init__.py
```diff
@@ -8,7 +8,7 @@
 def test_version_matches():
     dist = pkg_resources.get_distribution('ics')
     print(repr(dist), dist.__dict__, sys.path, ics.__path__)
-    assert len(ics.__path__) == 1
+    assert len(ics.__path__) is 1
```
C4ptainCrunch, line 11: Comparing integers with the identity is a no-go. It might work on the current implementation of  cPython with small numbers might work for now but does not work for bigger integer and is not guaranteed to work over time or with other implementation of Python (think pypy)

https://stackoverflow.com/a/306353

C4ptainCrunch (changes requested):

tests/__init__.py
pnijhara: Oh! I agree with this!
I am reverting this change.

N-Coder:
Hey, thanks for hinting us at that tool and providing the appropriate configuration.
Unfortunately, I'm not quite happy with the code changes you propose in this PR and some of the further changes proposed on the linked website.
IMHO an else-branch after a return is not an Anti-Pattern (as suggested [43 times](https://deepsource.io/gh/pnijhara/ics.py/issue/PYL-R1705/occurrences) on the website), but can in many cases improve readability, hinting you at the fact that the following code is not always executed and the if actually branches between two distinct code paths.
Also, instantiating empty tuples and dicts via their explicit constructor is preferable to me, as the syntactic sugar versions `()` and `{}` can be confusing regarding the actual data type they yield - e.g. a dict or a set?
Also some of the checks seem bogus - e.g. the unused import of BinaryConverter being reported [multiple times](https://deepsource.io/gh/pnijhara/ics.py/issue/PYL-W0611/occurrences), where probably it should instead also report the further imports following on the same line. Actually, all those imports are important as they ensure that those classes are loaded and registered, but that isn't caught by the tool.
So we'd probably still need some configuration to get really useful reports out of this.

One further issue is that we already have quite some checkers running, for which it is also not always easy to keep them satisfied, but at least they allow you to run the checks locally before committing. As far as I see you can't run those checks locally without committing and pushing that to github, right?

Maybe this is something we could use for the finishing touches of v0.8, but right now I'd prefer to spend my time on actually pushing that version forward instead of having another checker whose false-positives you have to somehow work around.

pnijhara:
> Maybe this is something we could use for the finishing touches of v0.8, but right now I'd prefer to spend my time on actually pushing that version forward instead of having another checker whose false-positives you have to somehow workaround.

If you find any issue as false positive you can simply mark that as "ignore false positive" in DeepSource dashboard. DeepSource takes false positives very seriously - so you don't need to somehow workaround.

Also, comparing to other checkers DeepSource has [Transformers](https://discuss.deepsource.io/t/transformers-automatically-format-all-incoming-code/123) which can Transform the whole PR with a commit. Along with this many issues can be automatically fixed by the help of DeepSource autofix tool.

C4ptainCrunch:
I think nobody has the time to handle this and i don't think this will add much value to the project, so i'm closing this for the moment. If someone wants to volunteer some time handling this, please comment and we could re-open :)

Issue #261, opened 2020-10-14
Documentation example does not work


pokoli:
I'm tryting to create a calendar file and [export it to a file](https://icspy.readthedocs.io/en/stable/index.html#export-a-calendar-to-a-file). So after reading the documentation I'm using the following code:

```python
>>> from ics import Calendar, Event
>>> c = Calendar()
>>> e = Event()
>>> e.name = "My cool event"
>>> e.begin = '20140101 00:00:00'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/pokoli/.virtualenvs/vet/lib/python3.8/site-packages/ics-0.7-py3.8.egg/ics/event.py", line 155, in begin
    value = get_arrow(value)
  File "/home/pokoli/.virtualenvs/vet/lib/python3.8/site-packages/ics-0.7-py3.8.egg/ics/utils.py", line 196, in get_arrow
    return arrow.get(value)
  File "/home/pokoli/.virtualenvs/vet/lib/python3.8/site-packages/arrow-0.14.7-py3.8.egg/arrow/api.py", line 21, in get
    return _factory.get(*args, **kwargs)
  File "/home/pokoli/.virtualenvs/vet/lib/python3.8/site-packages/arrow-0.14.7-py3.8.egg/arrow/factory.py", line 207, in get
    dt = parser.DateTimeParser(locale).parse_iso(arg)
  File "/home/pokoli/.virtualenvs/vet/lib/python3.8/site-packages/arrow-0.14.7-py3.8.egg/arrow/parser.py", line 131, in parse_iso
    return self._parse_multiformat(string, formats)
  File "/home/pokoli/.virtualenvs/vet/lib/python3.8/site-packages/arrow-0.14.7-py3.8.egg/arrow/parser.py", line 318, in _parse_multiformat
    raise ParserError(
arrow.parser.ParserError: Could not match input to any of ['YYYY-MM-DD HH:mm:ss'] on '20140101 00:00:00'
>>> c.events.add(e)
>>> e.begin = '2014-01-01 00:00:00'
>>> c.events.add(e)
>>> with open('my.ics', 'w') as f:
...     f.write(c)
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
TypeError: write() argument must be str, not Calendar
>>> 

```
So figured myself how to fix the data format but I do not now how to write the calendar to the file.

If someone gives me some advise I will create a PR to improve the docs. 

Thanks in advance

[...]

PR #262, opened 2020-10-14
Fix documentation quickstart example

pokoli:
As explained on https://github.com/C4ptainCrunch/ics.py/issues/261 here is the fix of the quickstart example to match the latest ics version.

PR #263, opened 2020-10-19
Fix documentation quickstart example

pokoli:


pokoli:
Fixes #261 following discussion from #262

aureooms (approved):

PR #291, opened 2021-08-05
Add py.typed file

whtsky:
Packages need to have a `py.typed` file to tell type checks "I have type information!", otherwise type checks won't use type informations provided in the package.

See:
https://www.python.org/dev/peps/pep-0561/#packaging-type-information
https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages
https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/


N-Coder:
Thanks for providing this! Did you check that poetry actually includes the file in the right place in the distribution?

whtsky:
> Did you check that poetry actually includes the file in the right place in the distribution?

Yes.
图片

PR #300, opened 2021-09-30
Don't normalize all-day events

rkdarst:
While testing normalization, I found that an all-day event will break `Calendar.normalize()`: `ValueError: can't replace timezone of all-day event`  This is latest default branch:

```
  File "yaml2ics/yaml2ics.py", line 119, in <module>
    calendar = files_to_calendar(files)
  File "yaml2ics/yaml2ics.py", line 103, in files_to_calendar
    calendar.normalize(dateutil.tz.gettz(tzname),
  File "/ics/icalendar.py", line 121, in normalize
    self.events = [normalization.normalize(e) for e in self.events]
  File "/ics/icalendar.py", line 121, in <listcomp>
    self.events = [normalization.normalize(e) for e in self.events]
  File "/ics/timespan.py", line 95, in normalize
    return action(value, replacement)
  File /ics/timespan.py", line 72, in replace_timezone
    value.replace_timezone(tzinfo=tzinfo)
  File "/ics/event.py", line 165, in replace_timezone
    self.timespan = self.timespan.replace_timezone(tzinfo)
  File "ics/timespan.py", line 143, in replace_timezone
    raise ValueError("can't replace timezone of all-day event")
```

I made a minimal change which might fix it (but it hasn't been tested), and I definitely don't know enough to say if this is even a correct or complete fix.  I would not consider this a fix but an issue with a diff attached (so it might be better to discard this and make a proper fix).

Thanks for working on this package, it has been helpful!


src/ics/icalendar.py
```diff
@@ -118,7 +118,7 @@ def normalize(self, normalization, *args, **kwargs):
                 raise ValueError("can't pass args or kwargs when a complete Normalization is given")
         else:
             normalization = Normalization(normalization, *args, **kwargs)
-        self.events = [normalization.normalize(e) for e in self.events]
+        self.events = [e if e.all_day else normalization.normalize(e) for e in self.events]
         self.todos = [normalization.normalize(e) for e in self.todos]
```
rkdarst, line 122: In particular, I am not sure if TODOs have times and could have the same all-day event+timezone normalization issue...

N-Coder:
Yeah, I guess normalization is even more bleeding-edge than some of the rest of the code...
I introduced this functionality to be able to compare floating (timezone-naive, i.e. with `tzinfo = None`) events with timezone-aware events, which is [surprisingly tricky](https://github.com/ics-py/ics-py/blob/main/doc/event-cmp.rst#comparing-datetimes-with-and-without-timezones). It seems that I didn't have all-day events in mind there, because those force `tzinfo` to be `None`. Factoring those in might make things even more tricky, because an all-day event on 5th of October would start at different times, depending on which timezone you are comparing against...
As a quick fix your code should work, but I'd prefer to later properly factor this into the `normalize` function itself, once I've come up with a way how this should work with comparisons and `normalize` also being used for timezone conversion. Could you apply the same fix for todos one line below and maybe also include some test case(s) (at least a very small one preventing regressions like [these here](https://github.com/ics-py/ics-py/blob/main/doc/event-cmp.rst#comparing-datetimes-with-and-without-timezones))?

rkdarst:
I've now pushed changes doing the conversion for todos, and added a test in `tests/normalization.py`.  The test fails when I revert the normalization improvement, so I hope it works... and I'm not sure if the test even covers what you asked for.  No promises, I'm not an expert here.

I agree that this fix isn't the best, but also, I'm not sure what the semantics of an "all-day event" even means.  To me, it doesn't imply a specific period of time, but I'm no expert and from the ics point of view, it doesn't even matter.  Good luck figuring out consistent semantics!

FWIW, I was using normalization to convert a calendar of naive datetimes into aware ones of a certain timezone, so that the timezone could be specified once, not for every event.  This was a somewhat thin mapping from a yaml file, so it was worth not having to copy-and-paste timezone info to each event.

PR #301, opened 2021-10-07
Drop Python 3.6 and add support for 3.10

C4ptainCrunch:
Python 3.10 is out since a few days, time to officially support it 🎉
Python 3.6 [will be EOL](https://www.python.org/dev/peps/pep-0494/#and-beyond-schedule) by the end of the year (so probably before the next ics release) so i'm suggesting dropping active support too.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "pypy-3.7", "pypy-3.8"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "pypy-3.7"]
 
     steps:
     - uses: actions/checkout@master
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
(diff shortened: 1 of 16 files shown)


C4ptainCrunch:
It looks like mypy had an upgrade since the last time the checks were ran and it breaks on 3.10 with the new mypy version.
I'll make a PR to 1) fix the error mypy is complaining about 2) pin mypy to a fixed version.

N-Coder:
Hey, great to see you back on here and thanks for merging!

1\) [`src/ics/converter/base.py:238:`](https://github.com/ics-py/ics-py/blob/main/src/ics/converter/base.py#L238) [`error: Need type annotation for "generic_vars"  [var-annotated]`](https://github.com/ics-py/ics-py/runs/4704239179?check_suite_focus=true) should be fixable by adding a `Tuple[Type]` annotation there.

2\) I'm not sure whether we should pin the mypy version, getting warnings about further potential issues continously doesn't sound too bad. This way we can fix these issues in small steps whenever they come up - altough that would probably usually be out of the current context. Otherwise we might get huge batches of mypy problems whenever we try to update the version, that is concentrated in one place.

C4ptainCrunch:
Hi 👋 

I'm back from 3 months vacation in another hemisphere, i'll try to contribute a bit more now that i'm back !
1) Thanks it was almost that and saved me a bit of time
2) I agree, let's occasionally break the build but never have a huge backlog of versions to upgrade 👍 

The PR is over there: https://github.com/ics-py/ics-py/pull/309

PR #303, opened 2021-11-07
Include ics.py version in PRODID field

N-Coder:
This PR will include the version of ics.py in the PRODID field, which currently defaults to `ics.py - http://git.io/lLljaA`, so that an empty calendar will now look like this:
```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:ics.py 0.8.0-dev - http://git.io/lLljaA
END:VCALENDAR
```
This should help with identifying the affected versions of bug reports. One could also include further system information for debugging such as the python interpreter and version and the OS, although this might be a problem with privacy (and possibly security, if we report a vulnerable version). Furthermore, including more information in the PRODID also makes deterministic outputs harder to get, e.g. see the updated bumpversion config. So I'm not 100% sure we should do this at all, but using only the version might be a good compromise between debugging information and reproducability.

### DIFF
--- a/src/ics/icalendar.py
+++ b/src/ics/icalendar.py
@@ -38,7 +38,7 @@ class Calendar(CalendarAttrs):
 
     NAME = "VCALENDAR"
     DEFAULT_VERSION: ClassVar[str] = "2.0"
-    DEFAULT_PRODID: ClassVar[str] = "ics.py - http://git.io/lLljaA"
+    DEFAULT_PRODID: ClassVar[str] = "ics.py 0.8.0-dev - http://git.io/lLljaA"
 
     def __init__(
             self,
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -11,7 +11,11 @@ serialize =
 
 [bumpversion:file:src/ics/__init__.py]
 
+[bumpversion:file:src/ics/icalendar.py]
+
 [bumpversion:file:doc/event.rst]
+
+[bumpversion:file:doc/timezone.rst]
 
 [bumpversion:file:doc/conf.py]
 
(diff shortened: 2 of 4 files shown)


make-github-pseudonymous-again (approved):

C4ptainCrunch:
This seems like a great idea ! I'll merge as soon as `main` is fixed :)

src/ics/icalendar.py
```diff
@@ -38,7 +38,7 @@ class Calendar(CalendarAttrs):
 
     NAME = "VCALENDAR"
     DEFAULT_VERSION: ClassVar[str] = "2.0"
-    DEFAULT_PRODID: ClassVar[str] = "ics.py - http://git.io/lLljaA"
+    DEFAULT_PRODID: ClassVar[str] = "ics.py 0.8.0-dev - http://git.io/lLljaA"
```
amureki, line 41: Greetings, friends!

I have a tiny question/suggestion here - I wonder if this field should be dynamic, based on `__version__` as well? To not rework it with every new version of the package...

N-Coder: bumpversion will take care of updating this value automatically, just as it does for all [other files](https://github.com/ics-py/ics-py/pull/303/files#diff-9ea6e1e3dde6d4a7e08c7c88eceed69ca745d0d2c779f8f85219b22266efff7f) where the version is hardcoded. We could also only have the version explicitly set in one place and then insert it dynamically at runtime in all other places, but that would require further code and runtime overhead and not really bring any benefit.

amureki: @N-Coder awesome, thanks for clarifications! Good to know about `bumpversion` 👍

PR #312, opened 2022-02-04
ENH: add support for calendar name

### LINKED ISSUE #307: Consider adding fields from RFC-7986
RFC-7986 allows for calendar proprties such as [https://icalendar.org/New-Properties-for-iCalendar-RFC-7986/5-1-name-property.html][a name].

Would adding this be in scope for the project?  I'm happy to make a PR.

tupui:
Closes #307

* Add `NAME` and `X-WR-CALNAME` support.
* `X-WR-CALNAME` is converted to `NAME` and `X-WR-CALNAME` is added in extras.
* Default name is used when the field is not present.

### DIFF
--- a/src/ics/converter/types/calendar.py
+++ b/src/ics/converter/types/calendar.py
@@ -19,7 +19,8 @@ class CalendarMeta(ComponentMeta):
 
     def find_converters(self):
         return sort_converters(itertools.chain(
-            super(CalendarMeta, self).find_converters(), (CalendarTimezoneConverter(),)
+            super(CalendarMeta, self).find_converters(),
+                (CalendarTimezoneConverter(), CalendarNameConverter())
         ))
 
     def _populate_attrs(self, instance: Component, container: Container, context: ContextDict):
@@ -60,6 +61,26 @@ class CalendarTimezoneConverter(GenericConverter):
     def serialize(self, component: Component, output: Container, context: ContextDict):
         # store the place where we should insert all the timezones
         context["VTIMEZONES_AFTER"] = len(output)
+
+
+class CalendarNameConverter(GenericConverter):
+    @property
+    def priority(self) -> int:
+        return 1100
+
+    @property
+    def filter_ics_names(self) -> List[str]:
+        return ["NAME", "X-WR-CALNAME"]
+
+    def populate(self, component: Component, item: ContainerItem, context: ContextDict) -> bool:
+        if item.name == "NAME":
+            return True
+        if item.name == "X-WR-CALNAME":
+            component.name = item.value
+            return False
+
+    def serialize(self, component: Component, output: Container, context: ContextDict):
+        pass
 
 
 ComponentMeta.BY_TYPE[Calendar] = CalendarMeta(Calendar)
--- a/src/ics/icalendar.py
+++ b/src/ics/icalendar.py
@@ -14,6 +14,7 @@ from ics.todo import Todo
 
 @attr.s
 class CalendarAttrs(Component):
+    name: str = attr.ib(validator=instance_of(str), metadata={"ics_priority": 1100})
     version: str = attr.ib(validator=instance_of(str), metadata={"ics_priority": 1000})  # default set by Calendar.DEFAULT_VERSION
     prodid: str = attr.ib(validator=instance_of(str), metadata={"ics_priority": 900})  # default set by Calendar.DEFAULT_PRODID
     scale: Optional[str] = attr.ib(default=None, metadata={"ics_priority": 800})
@@ -37,6 +38,7 @@ class Calendar(CalendarAttrs):
     """
 
     NAME = "VCALENDAR"
+    DEFAULT_CALNAME = "Calendar"
     DEFAULT_VERSION: ClassVar[str] = "2.0"
     DEFAULT_PRODID: ClassVar[str] = "ics.py 0.8.0-dev - http://git.io/lLljaA"
 
@@ -60,6 +62,7 @@ class Calendar(CalendarAttrs):
             events = tuple()
         if todos is None:
             todos = tuple()
+        kwargs.setdefault("name", self.DEFAULT_CALNAME)
         kwargs.setdefault("version", self.DEFAULT_VERSION)
         kwargs.setdefault("prodid", creator if creator is not None else self.DEFAULT_PRODID)
         super(Calendar, self).__init__(events=events, todos=todos, **kwargs)  # type: ignore[arg-type]
--- a/tests/test_calendar.py
+++ b/tests/test_calendar.py
@@ -1,0 +1,18 @@
+from ics import Calendar
+
+lines = [
+    "BEGIN:VCALENDAR",
+    "VERSION:2.0",
+    "PRODID:-//Apple Inc.//Mac OS X 10.13.5//EN",
+    "CALSCALE:GREGORIAN",
+    "BEGIN:VEVENT",
+    r"DESCRIPTION:Title: My event",
+    "END:VEVENT",
+    "END:VCALENDAR",
+]
+
+
+def test_cal_name():
+    calendar = Calendar(lines)
+
+    assert calendar.name == Calendar.DEFAULT_CALNAME


tupui:
Right now it fails to register. Although I can see that `CalendarNameConverter.populate` was called and modified `component`. It then fails when calling the `AttributeValueConverter.post_populate`. And I am stuck for now. Can anyone help?

```python-traceback
    def post_populate(self, component: Component, context: ContextDict):
        if self.is_required and not context[(self, "current_value_count")]:
>           raise ValueError("attribute %s is required but got no value" % self.ics_name)
E           ValueError: attribute NAME is required but got no value
```

tupui:
@N-Coder @C4ptainCrunch could you have a look? 😃

C4ptainCrunch:
Hi @tupui 👋 

Sorry for the delay. First, thanks a lot for taking the time to submit a PR, you rock 🤘 

I did not have a lot of time to look at your PR yet but i do have a question: Why do you add a default value for the calendar name ? Shouldn't we just keep it null if we did not get any data (and not writing a `NAME` property when serializing it back)

tupui:
> Hi @tupui 👋
> 
> Sorry for the delay. First, thanks a lot for taking the time to submit a PR, you rock 🤘

No worries!

> I did not have a lot of time to look at your PR yet but i do have a question: Why do you add a default value for the calendar name ? Shouldn't we just keep it null if we did not get any data (and not writing a `NAME` property when serializing it back)

I thought a calendar should be named, but no problem to make this completely optional and not present if not specified.

> UPDATE: I had a closer look and fiddled a bit with your code. My changes are available in [8761dfc](https://github.com/ics-py/ics-py/commit/8761dfc9e1f06b13d921e5a1dca419b71e6e17ec) if you want. I hope it unsticks you :)

Oh thanks! I will apply your changes then 😃

src/ics/converter/types/calendar.py
```diff
@@ -62,4 +63,24 @@ def serialize(self, component: Component, output: Container, context: ContextDic
         context["VTIMEZONES_AFTER"] = len(output)
 
 
+class CalendarNameConverter(GenericConverter):
```
tupui, line 66: I suppose we can then remove all changes to this file then?

Sorry I did not get the `optional` validator. I had trouble to understand how these were working together.

tests/test_calendar.py
```diff
@@ -0,0 +1,49 @@
...
+    assert calendar.name == "a name"
+
+def test_cal_x_name():
+    calendar = Calendar(cal2)
+    assert calendar.name == "a xname"
+
+def test_cal_both_names():
+    calendar = Calendar(cal3)
+    # FIXME: what behavior should we expect here?
+    # FIXME: In both cases, we need a test and we need to document it.
```
tupui, line 44: Since the RFC is only referring to `NAME`, I suppose one could say that if both fields are present, it should take NAME instead of X-WR-CALNAME. This is what I've done now.

C4ptainCrunch:
Great ! This looks in good shape to be merged :tada:

But first, could you :
* Add yourself to `AUTHORS.rst`
* Reference this change in `CHANGELOG.rst`
* Add documentation about the feature (and insist on the `NAME` vs `X-WR-CALNAME` precedence as well as the behaviour when `X-WR-CALNAME` is in the input and we serialize the output with `NAME`)

src/ics/converter/types/calendar.py
```diff
@@ -19,7 +19,8 @@ class CalendarMeta(ComponentMeta):
```
C4ptainCrunch, line 19: Could you update the docstring above explaining [this](https://github.com/ics-py/ics-py/pull/312/files/bafba59bb873105f76798105343878d43659e2a8#r805589474) ?

tupui: Done

N-Coder (changes requested):
Thanks a lot for making the effort to get your way around my Converter system and creating this PR - and sorry for taking my time to reply! Unfortunately, your code doesn't quite do what you intended, see my reviews. I see this as another argument for the Converter system being in desperate need of some refactoring and especially bigger simplifications. I guess the simplest way forward here would be using the following implementation for your `propagate`, adding handling of conflicting specifications later on once the Converter system makes that easier to do:
```python
     def populate(self, component: Component, item: ContainerItem, context: ContextDict) -> bool:
            component.name = item.value
            return False
```
One thing that should be noted: with this PR, the library will now [throw an exception](https://github.com/ics-py/ics-py/blob/main/src/ics/converter/value.py#L67) when it sees multiple values for name, e.g. because they were localized, while before the multiple values could be stored to the extra properties just fine. This the second big "construction site", where the library is too strict in many cases. But I guess supporting one one name for now is okay.

src/ics/converter/types/calendar.py
```diff
@@ -62,4 +68,27 @@ def serialize(self, component: Component, output: Container, context: ContextDic
...
+
+    def populate(self, component: Component, item: ContainerItem, context: ContextDict) -> bool:
+        if item.name == "NAME":
+            return True
+        if item.name == "X-WR-CALNAME":
+            if component.name is not None:
+                return False
+
+            component.name = item.value
+            return False
```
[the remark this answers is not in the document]
tupui: Actually I think we do want that because both google and apple do not handle NAME but only X-WR-CALNAME. I tested this locally and if you have NAME it does not care. It's unfortunate that they do not follow the RFC...

src/ics/icalendar.py
```diff
@@ -18,6 +18,7 @@ class CalendarAttrs(Component):
     prodid: str = attr.ib(validator=instance_of(str), metadata={"ics_priority": 900})  # default set by Calendar.DEFAULT_PRODID
     scale: Optional[str] = attr.ib(default=None, metadata={"ics_priority": 800})
     method: Optional[str] = attr.ib(default=None, metadata={"ics_priority": 700})
+    name: Optional[str] = attr.ib(default=None, validator=v_optional(instance_of(str)), metadata={"ics_priority": 1100})
```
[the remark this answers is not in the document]
tupui: Done

tupui (commented):
As I said bellow, I just came across the fact that we would need X-WR-CALNAME for practical reason.

Let me know what should be the behaviour. I personally would go with providing both fields (either another attribute or in extra) and NAME would take precedence over X-WR-CALNAME to follow a bit more the RFC.

src/ics/converter/types/calendar.py
```diff
@@ -62,4 +68,27 @@ def serialize(self, component: Component, output: Container, context: ContextDic
...
 
 
+class CalendarNameConverter(GenericConverter):
+    @property
+    def priority(self) -> int:
+        return 1100
+
+    @property
+    def filter_ics_names(self) -> List[str]:
+        return ["NAME", "X-WR-CALNAME"]
```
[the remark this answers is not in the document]
tupui: I suppose this will depend on what we do about NAME/X-WR-CALNAME

N-Coder:
> As I said bellow, I just came across the fact that we would need X-WR-CALNAME for practical reason.

I guess the best way to support this would be parsing both values into the same field `name`, and allowing the user to select a compatibility level for serialization (i.e. RFC5545, RFC7986, Google, Outlook, ...) that would then decide whether the value is written out as NAME or X-WR-CALNAME.

Unfortunately, we don't have that yet and it only adds to the complexity of refactoring the (de-)serialization system, so it will probably also take a while. In the meantime, I'd say the sanest thing to do is add a `name` field that only maps to `NAME` in the ics and let `X-WR-CALNAME` go entirely separate via the `extras`. Adding a comment to the `name` field about Google and `X-WR-CALNAME` and not doing any incomplete translation between both properties would then probably cause the least suprises for users.

tupui:
> Unfortunately, we don't have that yet and it only adds to the complexity of refactoring the (de-)serialization system, so it will probably also take a while. In the meantime, I'd say the sanest thing to do is add a `name` field that only maps to `NAME` in the ics and let `X-WR-CALNAME` go entirely separate via the `extras`. Adding a comment to the `name` field about Google and `X-WR-CALNAME` and not doing any incomplete translation between both properties would then probably cause the least suprises for users.

This is the current behaviour implemented in this PR no?

N-Coder:
If you remove the added `CalendarNameConverter` by reverting the changes to `src/ics/converter/types/calendar.py` that should be it. Then we simply add the name field and letting its Converter be autogenerated without ever touching the `X-WR-CALNAME`.

src/ics/icalendar.py
```diff
@@ -18,6 +18,7 @@ class CalendarAttrs(Component):
     prodid: str = attr.ib(validator=instance_of(str), metadata={"ics_priority": 900})  # default set by Calendar.DEFAULT_PRODID
     scale: Optional[str] = attr.ib(default=None, metadata={"ics_priority": 800})
     method: Optional[str] = attr.ib(default=None, metadata={"ics_priority": 700})
+    name: Optional[str] = attr.ib(default=None, validator=v_optional(instance_of(str)), metadata={"ics_priority": 600})
     # CalendarTimezoneConverter has priority 600
```
N-Coder, line 22: Please also decrement the priority of `CalendarTimezoneConverter` and update the comment here so that we get a consistent ordering with the name before the timezones.

src/ics/converter/types/calendar.py
N-Coder: > I suppose we can then remove all changes to this file then?

yes, please do so

tupui: But then we don't have `X-WR-CALNAME` at all if we do this. Shouldn't we try to always add `X-WR-CALNAME`?

tupui: I would have this in `CalendarNameConverter.populate` 

```
if item.name == "NAME":
    component.extra_params["X-WR-CALNAME"] = item.value
    return True
```

tupui: I can make the tests pass, but I am probably not doing it right, you're probably better of doing this yourself at this point 😄

C4ptainCrunch:
I'm closing this due to the previous comment. Anybody can reopen if they have time to work on this :)

stefanv:
This felt like an important issue. We can work around it, as we've been doing, but calendars without names are much less useful for importing on eg Google Calendar. If the PR is closed, shouldn't an issue at least be opened?

C4ptainCrunch:
@stefanv You are totally right ! I opened #375 to track this :)

PR #318, opened 2022-03-07
add warnings about breaking changes to Calendar.__str__ and __iter__ in v0.7(.1)

N-Coder:
In 0.8, there are some breaking changes to `str(calendar)` which can cause data loss if users update ics carelessly, so we should have at least one release that warns of this (see [here](https://github.com/ics-py/ics-py/issues/245#issuecomment-1059507609)).
As a stable 0.8 release is still somewhat in the future and we have some not-yet released fixes for version 0.7 lying around, my suggestion would be to add these warnings in a v0.7.1 release and push out all those fixes very soon (we could've actually done this in 2020 to give users even more time to prepare :see_no_evil:).

 I suggest that the implicit serialization in `Calendar.__iter__` should go in version 0.8, while the changed behaviour of `Calendar.__str__` is deferred to v0.9. For `__str__` in version 0.8, I'll keep the old serialization and the warning but also provide an easy opt-in to get the nicer future behaviour and thus suppress the warning.

@C4ptainCrunch if this sounds okay to you, could you merge this PR and then push a 0.7.1 release to PyPi?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -51,7 +51,7 @@ Quickstart
     c.events
     # [<Event 'My cool event' begin:2014-01-01 00:00:00 end:2014-01-01 00:00:01>]
     with open('my.ics', 'w') as my_file:
-        my_file.writelines(c)
+        my_file.writelines(c.serialize_iter())
     # and it's done !
 
 More examples are available in the
--- a/doc/examples.rst
+++ b/doc/examples.rst
@@ -42,9 +42,9 @@ Export a Calendar to a file
 .. code-block:: python
 
     with open('my.ics', 'w') as f:
-        f.writelines(c)
+        f.writelines(c.serialize_iter())
     # And it's done !
 
     # iCalendar-formatted data is also available in a string
-    str(c)
+    c.serialize()
     # 'BEGIN:VCALENDAR\nPRODID:...
--- a/ics/component.py
+++ b/ics/component.py
@@ -1,6 +1,6 @@
 import warnings
 from collections import namedtuple
-from typing import Any, Dict, Tuple
+from typing import Any, Dict, Tuple, Iterable
 
 from ics.grammar.parse import Container
 from .utils import get_lines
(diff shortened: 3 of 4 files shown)


N-Coder:
Making the v0.7.1 release will also close #244.

C4ptainCrunch:
Hey 👋 

Sorry, i don't have a lot of time at the moment. I'd like to suggest something else to fix this .__str__ and __iter__ problem.
I'll try to write my thoughts in the following days if i have some spare time :)

N-Coder:
@C4ptainCrunch any news here? We should make sure to have enough time between this and v0.8 and I don't see any big alternatives. Could you at least give us an idea or should we move ahead with merging this?

C4ptainCrunch:
Hi, sorry to be late again, it's not easy to find time these days 🙈 

First, i'm OK to drop `__iter__` (with some advance warning), it was a hack and i think it's not a very well known pattern elsewhere.
But for `__repr__` and `__str__` i have a different opinion.

To me, `repr` should be something small, readable that quickly helps the developer **when working in the shell**. 
Right now with 0.8, when using the example in the docs that [reads a calendar from the internet](https://icspy.readthedocs.io/en/stable/#import-a-calendar-from-a-file), typing `Calendar(requests.get(url).text)` in a shell outputs 	100k+ lines and floods the shell.

And `__str__` is sometimes something nice to print in a UI or sometimes just not defined (and falls back to `__repr__`)

For example, if we look at requests we have this:
```python
r = requests.get('http://exemple.com')
r # <Response [200]>
str(r) # <Response [200]>
```

And for Django it's kind of the same where the `repr` for a model looks like `<Model: Model object (1234)>` by default. And Django encourages the developer to override `__str__` to show something nice that will be printed in the Django admin.

Looking at numpy, it's also kind of the same. Only that `__str__` is implemented and gives an even shorter output, but loosing some important information about the type.
```python
a = numpy.zeros((200, 200))
r  # array([[0., 0., 0., ..., 0., 0., 0.],
   #    [0., 0., 0., ..., 0., 0., 0.],
   #    [0., 0., 0., ..., 0., 0., 0.],
   #    ...,
   #    [0., 0., 0., ..., 0., 0., 0.],
   #    [0., 0., 0., ..., 0., 0., 0.],
   #    [0., 0., 0., ..., 0., 0., 0.]])

str(r) # [[0. 0. 0. ... 0. 0. 0.]
# [0. 0. 0. ... 0. 0. 0.]
# [0. 0. 0. ... 0. 0. 0.]
# ...
# [0. 0. 0. ... 0. 0. 0.]
# [0. 0. 0. ... 0. 0. 0.]
# [0. 0. 0. ... 0. 0. 0.]]
```

What i'm suggesting for `ics` would be to have `__repr__` output something like `<Event begin: 2020-02-20 20:20:00 end: 2020-02-20 20:20:00>` or equivalent to keep the "Pythonic" convention that almost everybody uses and keep a concise representation where the user can immediately graps what type and approximative content the variable has when manipulating in the shell.

But this leaves 2 open questions:
 * What do we do with `__str__` ? Do we keep a serialised output like 0.7 and before or do we remove it or something else ? To me keeping it has 2 advantages: no backwards compatibility problems and it's easy to work with as `f = open(...); f.write(calendar)` works out of the box.
 * What do we do with the big fat `repr` made by `attr` ? It's still useful to inspect the object in detail and maybe even copy/paste it in the shell to make a copy. Should we keep it but maybe somewhere in `.dump()` (or a better name)

C4ptainCrunch:
Ok, you are right, let's go for it 🚀  I'm really sorry to have blocked this for so long :/

PR #325, opened 2022-05-24
Add test coverage for timeline (from 35% to >90%)

allenporter:
Add simple test cases for the calendar timeline to increase test coverage for timeline (from 35% to >90% according to tox codecov)

The motivation is mostly about learning the code base, and this seemed like something useful to do in the process.

### DIFF
--- a/tests/timeline.py
+++ b/tests/timeline.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from datetime import datetime, timedelta, date
 
 from freezegun import freeze_time
--- a/tests/timeline.py
+++ b/tests/timeline.py
@@ -14,10 +14,10 @@ def calendar() -> Calendar:
     """Fixture calendar with all day events to use in tests."""
     cal = Calendar()
     cal.events.extend([
-        Event("second", datetime(2000, 2, 1, 12, 0)),
-        Event("fourth", datetime(2000, 4, 1, 12, 0)),
-        Event("third", datetime(2000, 3, 1, 12, 0)),
-        Event("first", datetime(2000, 1, 1, 12, 0)),
+        Event("second", date(2000, 2, 1), date(2000, 2, 2)),
+        Event("fourth", date(2000, 4, 1), date(2000, 4, 2)),
+        Event("third", date(2000, 3, 1), date(2000, 3, 2)),
+        Event("first", date(2000, 1, 1), date(2000, 1, 2))
     ])
     return cal
 
@@ -67,7 +67,7 @@ def test_start_after(calendar: Calendar) -> None:
     """Test chronological iteration starting at a specific time."""
     assert [
         e.summary
-        for e in calendar.timeline.start_after(date(2000, 2, 1))
+        for e in calendar.timeline.start_after(date(2000, 1, 1))
     ] == [ "second", "third", "fourth" ]
 
 
--- a/tests/timeline.py
+++ b/tests/timeline.py
@@ -1,10 +1,13 @@
 from datetime import datetime, timedelta, date
 
+import pytest
+
 from ics import Calendar, Event
 from ics.timezone import UTC
 
 
-def make_calendar():
+@pytest.fixture
+def calendar() -> Calendar():
     """Create a test calendar for use in tests."""
     cal = Calendar()
     cal.events.extend([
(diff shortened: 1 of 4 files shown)


tests/timeline.py
```diff
@@ -0,0 +1,148 @@
...
+    ] == ["event"]
+    assert [
+        e.summary for e in cal.timeline.at(datetime(2000, 1, 1, 12, 59))
+    ] == ["event"]
+    assert [
+        e.summary for e in cal.timeline.at(datetime(2000, 1, 1, 13, 0))
+    ] == []
+
+
+def test_now():
```
C4ptainCrunch, line 82: I'm not a huge fan of this test for 2 reasons:
* Using `datetime.now()` in tests usually breaks at some point, either because of DST, or something else
* You are not only testing `now()` but also `at()` in this test so if we break `at()` at some point, this test will fail too but `now()` would still be correct.

I see 2 options here:
1. Do not test this function, its a 1 line ( `return self.at(datetime.utcnow())`), what could go wrong ? 
2. Use something like [freezegun](https://github.com/spulec/freezegun) to fix the time, then mock  `at()` and verify the mock was indeed called with the frozen time you set.

If `now()` was a complicated function that we could mess up easily, i would go for option 2. But here, the test would be 10x longer than the actual function that is being tested and it would need a new dependency so i suggest to drop this test :)

tests/timeline.py
```diff
@@ -0,0 +1,148 @@
+from datetime import datetime, timedelta, date
+
+from ics import Calendar, Event
+from ics.timezone import UTC
+
+
+def make_calendar():
```
C4ptainCrunch, line 7: Could you make this a [pytest fixture](https://docs.pytest.org/en/6.2.x/fixture.html) please ?

C4ptainCrunch (commented):
Thanks @allenporter ! It is indeed useful :)

C4ptainCrunch:
Also, could you pull `main` into your branch ? I just made a fix for GitHub Actions so the tests will run on pull requests from forks :)

tests/timeline.py
allenporter: 👍🏼

tests/timeline.py
allenporter: I went with #2 and i think combined with the move to use fixtures its a little less offensive.  That is, if we use fixtures for all the datetime based tests then this test doesn't really stick out as being overly complex use with freezegun. I also added a test for `today` while I was here.

If you have a look and still think its not worth having, then happy to drop, just let me know what you think.

allenporter:
> Also, could you pull `main` into your branch ? I just made a fix for GitHub Actions so the tests will run on pull requests from forks :)

Done.

allenporter:
> > Also, could you pull `main` into your branch ? I just made a fix for GitHub Actions so the tests will run on pull requests from forks :)
> 
> Done.

This caught a few typing issues that i'll resolve now.

tests/timeline.py
```diff
@@ -0,0 +1,148 @@
+from __future__ import annotations
```
N-Coder, line 1: we can probably get rid of this

tox.ini
```diff
@@ -4,6 +4,7 @@ envlist = py37, py38, py39, py310, py311, pypy37, pypy38, pypy39, checks, docs
 
 [testenv]
 description = Run the pytest suite
+deps = freezegun>=1.2.1
```
N-Coder, line 7: this would probably be better configured in pyproject.toml amongst the testing extras ;)

tests/timeline.py
```diff
@@ -0,0 +1,148 @@
...
+from freezegun import freeze_time
+import pytest
+
+from ics import Calendar, Event
+from ics.timezone import UTC
+
+
+@pytest.fixture
+def calendar() -> Calendar:
+    """Fixture calendar with all day events to use in tests."""
```
N-Coder, line 14: I guess these are not really all-day events as they simply start at noon and end in the same instant. Would it be easier to mix different event types in the same Calendar (i.e. all-day, multi-day all-day, no specified end, fixed end or duration, maybe even with/without timezones?) and then thoroughly test that calendar?

Note: We could also infer that an Event should be all-day if its created from `date`s instead of `datetime`s, but currently the `ensure_datetime` [here](https://github.com/ics-py/ics-py/blob/e62d660/src/ics/event.py#L273) silently sets the time to midnight if its missing and ignores this distinction. Something like the following might fix that but needs further testing.
```python
if (begin is not None or end is not None or duration is not None) and "timespan" in kwargs:
    raise ValueError("can't specify explicit timespan together with any of begin, end or duration")
timespan = kwargs.pop("timespan", EventTimespan(ensure_datetime(begin), ensure_datetime(end), ensure_timedelta(duration)))
if isinstance(begin, date) and not timespan.get_effective_duration().seconds:  # TODO we should also drop microseconds to be able to ignore them here
    timespan = timespan.make_all_day()
super(Event, self).__init__(timespan, summary, *args, **kwargs)
```

N-Coder (approved):
Two minor things not related to the tests and a suggestion for testing even more cases in a single Calendar, but this is definitely a good start!

tests/timeline.py
allenporter: My impression is this is needed to use newer typing:

```
________________________________ ERROR collecting tests/samples.py _________________________________
import file mismatch:
imported module 'tests.samples' has this __file__ attribute:
  /home/allen/ics-py/tests/samples
which is not the same as the test file we want to collect:
  /home/allen/ics-py/tests/samples.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
________________________________ ERROR collecting tests/timeline.py ________________________________
tests/timeline.py:61: in <module>
    def test_on(calendar: Calendar, when: date | datetime, expected_events: list[str]) -> None:
E   TypeError: unsupported operand type(s) for |: 'type' and 'type'

```

tox.ini
allenporter: Thanks, first time using poetry. Moved.

tests/timeline.py
allenporter: Whoops, I had intended for these to be all day events with `date` as the input, but misread them when setting them up.

Let me update them to be what they were intended to be, then see what you think.

tests/timeline.py
N-Coder: Oh, this is the first future import I see that might actually make sense in modern python. This enables the deferred evaluation of annotations, so python no longer tries to interpret this annotation at runtime and instead simply stores it as string:
https://docs.python.org/3/library/__future__.html
I guess in this specific case the issue might be that the pipe for type unions is python 3.10 syntax and you're using an older python?
https://peps.python.org/pep-0604/
Anyways, it should be fine if you leave it in.

tests/timeline.py
N-Coder: Note that the new functionality regarding all-day events I proposed is not yet implemented, so this still generates normal events ranging from midnight to midnight. :sweat_smile:
Currently, you'd need to explicitly call `make_all_day` on an event, but I'll probably include the code above to prevent any confusion with this.

allenporter: I've added `make_all_day` calls for now, and agree the above snippet to make this transparent makes a lot of sense.

PR #326, opened 2022-05-29
Restore repr() to a small string used in the shell

C4ptainCrunch:
See https://github.com/ics-py/ics-py/pull/318#issuecomment-1140403955.

Also `str()` calls `serialize()` again.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -3,8 +3,12 @@ import sys
 import pkg_resources
 
 import ics
+import pytest
+
+ics_path = ics.__path__[0]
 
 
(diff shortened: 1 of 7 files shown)


tests/__init__.py
```diff
@@ -3,8 +3,12 @@
 import pkg_resources
 
 import ics
+import pytest
 
+ics_path = ics.__path__[0]
 
+
+@pytest.mark.skipif("/site-packages/" not in ics_path or "/src" in ics_path, reason="ics should be imported from package not from sources for testing")
```
N-Coder, line 11: why are you skipping this?

PR #332, opened 2022-05-29
Upgrade to Sphinx 5

C4ptainCrunch:


pyproject.toml
```diff
@@ -45,7 +45,8 @@ bump2version =               { version = "^1.0.0",     optional = true }
 tox =                        { version = "^3.15.0",    optional = true }
 
 # extra: docs
-sphinx =                     { version =  "<3",        optional = true }
+sphinx =                     { version =  ">=5,<6",        optional = true }
+Sphinx = "^5.0.0"
```
N-Coder, line 49: why specify this twice?

C4ptainCrunch: It's an error, but it's fixed in https://github.com/ics-py/ics-py/blob/main/pyproject.toml#L48
Again, i should not merge my own PRs, sorry.

PR #349, opened 2022-06-30
Unpin arrow

C4ptainCrunch:


N-Coder:
:+1: to merging this. I'd test it together with an updated test pypi release. If we do find any issues, it should be no problems to fix them in a further PR (or via direct commits to the branch as it isn't really highly frequented).

Issue #344, opened 2022-06-17
Code formatting


strobeflash:
> Formatting guidelines would be good too

As far as I'm aware we currently have neither code nor doc formatting guidelines, but these would definitely be good to have. For code we might want to switch to `black` at some point (I'm currently using the PyCharm formatter defaults), but for docs we probably just need to agree on something and stick to that. There's the default [rst](https://www.sphinx-doc.org/en/master/tutorial/describing-code.html) style, [numpy](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) and the [google](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings) style. A comparison of the three can be found further down the [napoleon](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) doc page. We already enabled that Sphinx plugin so any of the three styles should work. Personally, I guess I'd prefer the Google style for parameters and the like as it requires the least amount of superfluous formatting and is nicely readable on a screen that is more wide than tall. :wink: @C4ptainCrunch @tomschr opinions?

_Originally posted by @N-Coder in https://github.com/ics-py/ics-py/issues/340#issuecomment-1152075377_

strobeflash:
Decoupled from pull request.

Low priority item to discuss code formatting guidelines. If a decision in favor of `black` is made, the [recommended strategy](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html) is to do it in one massive commit and then use [.git-blame-ignore-revs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view), that is natively supported by github.

tomschr:
As far as I know, code formatting can be done with black and other tools. I've integrated it myself as a tox target in [another project](https://github.com/python-semver/python-semver/blob/master/tox.ini#L26-L30). This is quite straightforward.

Depending on your level of dedication to code formatting, you can integrate it as a GH Action, for example, for every pull request. That would make it not a one-time thing, but it will remind all developers about it (for better or worse).

Regarding doc strings, there is [docformatter](https://pypi.org/project/docformatter/) which checks against [PEP257](http://www.python.org/dev/peps/pep-0257/). This is only a rudimentary check for correct indentation. It doesn't check if you omitted an argument in your docstrings etc.

[...]

PR #351, opened 2022-07-05
Introduce auto-formatters

C4ptainCrunch:
See #344 

Adds pre-commit, black, isort, pyupgrade, blacken-docs.

Missing:

- [x] Update the (contributing) documentation
- [ ] GitHub Action (we can use https://github.com/pre-commit/action)

This PR will make a lot of noise: changing almost every file, sneaking into git blame, introducing conflicts in open PR, ...
So i would like to:

1. Wait until Monday 11th of July at least to give a chance for open PRs to be merged before the big-bang
2. Introduce as much auto-formatters as possible so that we don't create others PR like this in a few months.

Should we add some more auto-formatters ? Something like https://github.com/fsouza/autoflake8 maybe ? Do you have some other ideas ? Now is the time !


.pre-commit-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+    -   id: trailing-whitespace
+-   repo: https://github.com/asottile/pyupgrade
+    rev: v2.34.0
+    hooks:
+    -   id: pyupgrade
+        args: [--py37-plus]
+-   repo: https://github.com/pycqa/isort
+    rev: 5.10.1
+    hooks:
+        - id: isort
```
allenporter, line 15: You may find you also need `args: ["--profile", "black"]`

allenporter (commented):
These also come to mind:
- Yes, also flake8 (though don't know what autoflake8 is)
- pylint

.pre-commit-config.yaml
C4ptainCrunch: It's specified in `pyproject.toml` https://github.com/ics-py/ics-py/pull/351/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R62 so if somebody runs it from the cli or and IDE it still works

N-Coder:
Wow, this seems to be moving fast. Maybe the deadline is a little bit tight given that we haven't fully figured out what we want and haven't even tested it? Here are some of my considerations:
- The suggested hooks look good, although it seems we could also enable a bunch of further hooks that come built-in with pre-commit (e.g. check-ast, check-case-conflict, end-of-file-fixer, mixed-line-ending, trailing-whitespace). I don't think we need any auto-fixers like autoflake8, esp. not in pre-commit hooks. Plus mypy and flake8 are at times already hard enough to reign in, so I guess we're already good in terms of code analysis.
- We still need to agree on a doc comment style and tooling for supporting that in #344 
- We are already running flake8 and mypy via tox and also on every pushed commit in GitHub Actions. Which checks do we want to forcibly run locally on every commit, which ones do we want to run online in GitHub Actions / on-demand offline via tox? In my opinion, commit hooks should be sufficiently fast to not interrupt your workflow esp. when doing multiple small commits. So we should find a good split here -- also given that running the "optional" checks locally is just as easy (if you remember to do it) thanks to tox. Maybe the pre-commit hooks should just check formatting and the more heavy-weight things should be called directly?
- Also, there are few configuration options we might wanna think about. Having a screen that is more wide than tall, I usually prefer lines longer than 80 characters over a pyramid of indentations built before a line broken into five overly indented overly short parts. So, I'd prefer to have a [line length](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length) greater than 100, and anyways this configuration probably needs synchronization across tools.
- How should we proceed here? My suggestion would be - in parallel while trying to get as many PRs as possible merged - to already set up a PR where our selection of tools ran and where we can inspect the results. Once we agree that we are good to go, we can always re-run all the checks on the current master and then force-push and merge said PR.

allenporter:
> Wow, this seems to be moving fast. Maybe the deadline is a little bit tight given that we haven't fully figured out what we want and haven't even tested it? Here are some of my considerations:
> 
> * The suggested hooks look good, although it seems we could also enable a bunch of further hooks that come built-in with pre-commit (e.g. check-ast, check-case-conflict, end-of-file-fixer, mixed-line-ending, trailing-whitespace). 

A few of these seem to be covered already e.g. by black

> I don't think we need any auto-fixers like autoflake8, esp. not in pre-commit hooks. Plus mypy and flake8 are at times already hard enough to reign in, so I guess we're already good in terms of code analysis.
> * We still need to agree on a doc comment style and tooling for supporting that in [Code formatting #344](https://github.com/ics-py/ics-py/issues/344)

Seems like this can be addressed in a follow up PR.

> * We are already running flake8 and mypy via tox and also on every pushed commit in GitHub Actions. Which checks do we want to forcibly run locally on every commit, which ones do we want to run online in GitHub Actions / on-demand offline via tox? In my opinion, commit hooks should be sufficiently fast to not interrupt your workflow esp. when doing multiple small commits. So we should find a good split here -- also given that running the "optional" checks locally is just as easy (if you remember to do it) thanks to tox. Maybe the pre-commit hooks should just check formatting and the more heavy-weight things should be called directly?

Are there specific checks in this list that have performance concerns?

"if you remember to do it" is the issue -- keep in mind the time spent if you miss a check, and find out later in a github action. It's much less hostile to just enforce with a pre-commit hook.  This seems like an established best practice, and not really an issue in practice given the speed of most of these invoked from pre-commit.

> * Also, there are few configuration options we might wanna think about. Having a screen that is more wide than tall, I usually prefer lines longer than 80 characters over a pyramid of indentations built before a line broken into five overly indented overly short parts. So, I'd prefer to have a [line length](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length) greater than 100, and anyways this configuration probably needs synchronization across tools.
> * How should we proceed here? My suggestion would be - in parallel while trying to get as many PRs as possible merged - to already set up a PR where our selection of tools ran and where we can inspect the results. Once we agree that we are good to go, we can always re-run all the checks on the current master and then force-push and merge said PR.

C4ptainCrunch:
> Wow, this seems to be moving fast. Maybe the deadline is a little bit tight [..]

Indeed, i'm sometimes a bit impatient 🙈 Also, the target was more to merge the others PR so we are clear to make big changes, not to merge the black PR today.

> >  it seems we could also enable a bunch of further hooks that come built-in with pre-commit 
> 
> A few of these seem to be covered already e.g. by black

It could still be nice to enable check-case-conflict, end-of-file-fixer, mixed-line-ending, trailing-whitespace to fix the non-python files imo.

> Are there specific checks in this list that have performance concerns?

To me, mypy is a bit too slow to be enabled in the pre-commit , i think. (But i don't have a very strong opinion on this.)

> So, I'd prefer to have a [line length](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length) greater than 100

I'm also for longer line lengths in general, the time where we had very small screens is long behind us. (But again, not a strong opinion)

N-Coder:
Nice, thanks for running all these things! Do you recall how long the individual steps / the overall process took? The results look pretty good to me, especially because it seems that the limit is not strictly enforced for strings, where breaking up would really hinder readability. The only weird thing is how format strings are sometimes replaced by the "{}".format() notations, sometimes converted to f-strings and sometimes left as-is. As f-string were introduced in 3.6, they would be the way to go everywhere - do we have a way about warning when the other formats are used?

C4ptainCrunch:
After the first run (that installs the tools, etc), a full run takes 2.57s on my machine. But we also have to keep in mind that a full run is almost never done as by default, pre-commit runs only on modified files so it's much much faster.

```
$ time pre-commit run --all
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
check for case conflicts.................................................Passed
mixed line ending........................................................Passed
pyupgrade................................................................Passed
isort....................................................................Passed
black....................................................................Passed
blacken-docs.............................................................Passed
pre-commit run --all  2.57s user 0.66s system 323% cpu 1.000 total
```

> format strings are sometimes replaced by the "{}".format() notations, sometimes converted to f-strings and sometimes left as-is
This is weird, i'll have a look at the options of pyupgrade to try and force f-strings everywhere.

C4ptainCrunch:
The doc of pyupgrade [says](https://github.com/asottile/pyupgrade#f-strings) :
> pyupgrade is intentionally timid and will not create an f-string if it would make the expression longer or if the substitution parameters are sufficiently complicated (as this can decrease readability).

And i see no option to override this behaviour.

I had a look around the internet and it seems that there is another tool to do this: [flynt](https://github.com/ikamensh/flynt).
Here is an example of what it could output if we also add is as a pre-commit: https://github.com/ics-py/ics-py/commit/e9735ccfb215b095c713a905b764f7f30dc220c7

Note: what we could also do, is to run it once right now to get rid of the legacy and then rely on pyupgrade afterwards, this might be good enough.

jarrodmillman:
> Note: what we could also do, is to run it once right now to get rid of the legacy and then rely on pyupgrade afterwards, this might be good enough.

I've done that in several projects (e.g., https://github.com/networkx/networkx) and it worked well.

C4ptainCrunch:
It took a few commits to pass cleanly the tests but i'm happy to say that it's now merged and that we have a GitHub action enforcing the pre-commits !

PR #352, opened 2022-07-05
set minimal setuptools version to 60.1.1

strobeflash:
due to https://github.com/pypa/setuptools/issues/2980

couldnt find a more elegant solution than adding the deps three times.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,6 +4,8 @@ envlist = py37, py38, py39, py310, py311, pypy37, pypy38, pypy39, checks, docs
 
 [testenv]
 description = Run the pytest suite
+deps =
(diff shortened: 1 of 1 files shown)


N-Coder:
I plan to migrate to `hatch` anyways (details are for another PR), so maybe this won't be needed in the end?

strobeflash:
Could be. Until then tox will fail in quite an ungly way for anyone not meeting this requirement in their virtualenv, see https://pastebin.com/0DmwA4D6

strobeflash:
@N-Coder i just tested with the latest commits and this bug was fixed with the migration to hatch as you predicted. thank you!

Issue #348, opened 2022-06-29
Dependency on Arrow?


will-emmerson:
There is a vulnerability in Arrow (https://github.com/advisories/GHSA-cjw4-2w9r-r8mv) but I can't use a newer version because ics requires an older version:
```
ics==0.7
  - arrow [required: >=0.11,<0.15, installed: 0.14.7]
    - python-dateutil [required: Any, installed: 2.8.2]
      - six [required: >=1.5, installed: 1.16.0]
  - python-dateutil [required: Any, installed: 2.8.2]
    - six [required: >=1.5, installed: 1.16.0]
  - six [required: >1.5, installed: 1.16.0]
  - tatsu [required: >4.2, installed: 5.8.0]
```
I can't seem to find where Arrow is imported otherwise I would make a pull request.

N-Coder:
Arrow is no longer used in the currently in-dev version v0.8. We were about to make a new maintenance release of the previous version, see #345 and the `version-0.7` branch, so it might be a good time to fix that now. @C4ptainCrunch do you know why we cap Arrow?

C4ptainCrunch:
I have a vague souvenir that 0.15 had some breaking changes so that's why we caped it. But the [changelog](https://arrow.readthedocs.io/en/latest/releases.html#id26) doesn't ring a bell immediately .

C4ptainCrunch:
It looks like my memory was correct, we had some incompatibility but it was for an even earlier version of arrow: https://github.com/ics-py/ics-py/commit/542eda84cdc59f645873663462d7086e6babda13 so it might be ok. I'll try to upgrade and see if the CI turns red.

C4ptainCrunch:
Oh, i think that there is a misunderstanding. The CVE is for **Apache** Arrow (a columnar memory format) and not `arrow` (the date&time library) that we use. It seems like the GitHub CVE is incorrect.

C4ptainCrunch:
I have a PR: #349 that unpins arrow and that seems to pass the tests but we might want to check a bit more as i don't fully trust our test suite :D

C4ptainCrunch:
Here it is : https://test.pypi.org/project/ics/0.7.1.dev1/

N-Coder:
Just tested the release, also including some arrow-ish things, and fixed some warnings (see [b8df69c](https://github.com/ics-py/ics-py/commit/b8df69ccbcbeaa86fb99c97ab4aa74cec727d4cd)), now that release seems good to go for me! #345

PR #354, opened 2022-07-10
Move build system from poetry to hatch

N-Coder:
It seems that [hatch](https://hatch.pypa.io/latest/) is about to be PyPA's successor to setuptools and in my experience with other projects it's so much nicer to use (and esp. less-opinionated) than Poetry, so here's a PR for moving to it. The pyproject.toml file now looks a lot cleaner and also follows the corresponding PEPs more closely. You shouldn'y actually notice this change much (except if you are directly invoking `poetry`) as `pip install`, `python -m build`, `tox`,... nicely abstract across the build backend and automatically use the correct one.
- project and dependency declaration is now according to the standard
- optional-dependencies are declared nicely (including pinned tox versions for flake8 and mypy)
- upper version caps preferred by poetry have been removed (see also [here](https://github.com/ics-py/ics-py/issues/348#issuecomment-1170927688))
- removed lock file which we don't need

Also fixes #352.


C4ptainCrunch (approved):
I have a hard time following all the new packaging developments in Python right now (i already did not grasp poetry fully) so i trust you on this. Thank you for taking the time to modernise our setup 🚀 

(Small detail before merging, don't forget to update [the contributing docs](https://github.com/ics-py/ics-py/blob/33a8d31b89f32173085edc59edb3ac9157a182c1/CONTRIBUTING.rst#setting-up-the-development-environment) ;) )

strobeflash:
> (Small detail before merging, don't forget to update [the contributing docs](https://github.com/ics-py/ics-py/blob/33a8d31b89f32173085edc59edb3ac9157a182c1/CONTRIBUTING.rst#setting-up-the-development-environment) ;) )

See https://github.com/ics-py/ics-py/pull/357, does that fit the bill?

PR #364, opened 2022-09-18
README: replace e.summary with e.name

1t53c5v03:
Quickstart code doesn't work properly because the non-existent property `summary` of class `Event` is used instead of `name`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -45,7 +45,7 @@ Quickstart
 
  c = Calendar()
  e = Event()
- e.summary = "My cool event"
+ e.name = "My cool event"
  e.description = "A meaningful description"
  e.begin = datetime.fromisoformat("2022-06-06T12:05:23+02:00")
  e.end = datetime.fromisoformat("2022-06-06T13:05:23+02:00")


N-Coder:
Hey, I guess the best way would be to make one the alias for the other using properties and add a respective comment to the docs to not introduce further breaking changes and keep in sync with the standard (that says summary iirc) while also providing a more logical interface (users typically expect this value to be called "name"). I've got the respective changes already lying around together with a fix for the date-based all-day event constructor, but didn't get to pushing them yet. Will probably also still take two weeks or so as I'm still on holiday, sorry.

C4ptainCrunch (approved):

erikvanzijst:
Can we merge this? I ran into the same issue and was also just about to create a PR. It's been a year and a half and it's only a one line change.

N-Coder:
Did anyone check that this still works with the `master` branch and not only with what is on PyPI?

N-Coder:
`summary` is actually [correct](https://github.com/ics-py/ics-py/blob/main/src/ics/event.py#L274), so closing this

PR #368, opened 2022-10-12
Update README.rst

agn-7:
Add new API methods to the code example to avoid misleading the user

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -45,11 +45,11 @@ Quickstart
 
  c = Calendar()
  e = Event()
- e.summary = "My cool event"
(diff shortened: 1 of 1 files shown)


N-Coder:
Have you tried this with the code on the main branch? IIRC the example was already updated and switching to `set` semantics for `events` looks like going backwards...

C4ptainCrunch (dismissed):

C4ptainCrunch:
#364 does take care of the example and indeed we don't want to get back to the `set` for events. I'll close this since it's a duplicate.

PR #374, opened 2022-12-06
add a script to test ics files in bulk

strobeflash:


C4ptainCrunch:
I like this idea :)
Would you consider making it a command line script that ics provides ? Something like described there https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html
I guess it could also benefit from using argparse instead of `ICAL_DIRECTORY` :)

strobeflash:
I added argparse.
What are we trying to achieve with a command line script according to the documentation? To make it available as an executable to the user on pip install?

make-github-pseudonymous-again:
@strobeflash Thanks for the script! Why does the PR include changes other than the new script?

strobeflash:
@make-github-pseudonymous-again i had rebased against the wrong branch, should be fixed now. thanks for noticing.

PR #378, opened 2022-12-29
Fix example in `README.rst`

chrulm:
Seems like `c.events` got changed to a `Set[Event]`, not a `List[Event]`, so just updating the example here as well.


N-Coder:
It's actually the other way around, events used to be a set and got changed to a [list](https://github.com/ics-py/ics-py/blob/19884443f7bccea8a1c9f93d8f2a4b15321ac421/src/ics/icalendar.py#L27) (preserving any intended order of events). This version is only released as alpha to pypi, though. So by default you get the old version using set, while our GitHub frontpage already shows the new, updated version.

So things are correct as we are, we are just behind with the release. Thanks for bringing this up anyways!

PR #395, opened 2023-08-15
Drop EOL Python 3.7 and add Python 3.12

C4ptainCrunch:


C4ptainCrunch:
Tests are still failing due to our dependency to [lipsum](https://github.com/thanethomson/lipsum) using `pkg_resources`, i guess we will have to get rid of it first.

jarrodmillman (approved):

stefanv:
Great, this seems to do the trick! :pray:

PR #402, opened 2023-10-03
Fix example with extra ContentLine

s3bw:
There's no module called `ics.parse` I believe `ContentLine` has moved to the module `ics.contentline`.

It's imported in the `__init__.py` file so I've referenced it with `ics.ContentLine`.

### DIFF
--- a/src/ics/event.py
+++ b/src/ics/event.py
@@ -279,8 +279,13 @@ class Event(EventAttrs):
     a duration *or* end instant.
 
     Unsupported event attributes can be found in `event.extra`,
-    a :class:`ics.parse.Container`. You may add some by appending a
-    :class:`ics.parse.ContentLine` to `.extra`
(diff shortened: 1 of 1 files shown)


tamis-laan:
I'm on the latest and I can't import content line:

``` python
from ics import Calendar, ContentLine
```

When I print the type I get:
`<class 'ics.grammar.parse.ContentLine'>`

I installed the library with pip, version `0.7.2`

Also the docs have not been updated properly:
https://icspy.readthedocs.io/en/stable/advanced.html#custom-properties

> It will result in an event that will have a **ics.parse.Container** in .extra:

> Just add some Container or **ics.parse.ContentLine** and they will appear in the output too.

C4ptainCrunch:
Your fixes have been incorporated into 28a289cfc30b05106bad85a40c7af89733c81a48, thanks !

PR #408, opened 2023-11-26
Fix grammatical errors for readability

azh412:
There were some trivial grammatical errors in `icalendar.py` which would make the documentation for `parse_multiple` harder to read. There was also an incorrect article of speech in the class comment for `Calendar`, so that was fixed.


C4ptainCrunch:
Hello !
Thank you for your pull request :)
Could you please add a second commit adding yourself to https://github.com/ics-py/ics-py/blob/main/AUTHORS.rst (if it's something you wish)

azh412:
Great, thanks!

PR #422, opened 2024-11-06
Drop EOL Python 3.8 and add Python 3.13

edgarrmondragon:


N-Coder:
Thanks! I guess we need to fix the CI first and then we should be good to merge this.

Issue #393, opened 2023-08-06
README showcases wrong usage for appending to events


Marc-commits:
Appending to c.events throws an AttributeError.
The quickstart should read c.events.add(e).

#394

N-Coder:
The master branch has `events` replaced with a list instead of a set, so using `append` is correct for that version. Thanks for bringing this up still.

PR #430, opened 2025-02-23
Fix readme example (events is a set)

n8henrie:
Running the example gives `AttributeError(''set' object has no attribute 'append'')` -- fix to use `add`  as per https://icspy.readthedocs.io/en/stable/#create-a-new-calendar-and-add-events

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -54,7 +54,7 @@ Quickstart
 
  c = Calendar()
  e = Event()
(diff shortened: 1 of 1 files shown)


n8henrie:
Also it looks like event.summary is now `name`

N-Coder:
See #393

PR #442, opened 2026-09-18
Fix event transparency and restore CI validation

### LINKED ISSUE #434: transparent=False should give TRANSP:OPAQUE
In 0.8.0-dev0 , currently when I create an Event with transparent=False, it results in the ICS output containing TRANSPARENT:FALSE , whereas it should result in TRANSP:OPAQUE . Similarly, transparent=True results in TRANSPARENT:TRUE ,whereas it should result in TRANSP:TRANSPARENT . I believe this used to work in 0.7.x .

x0Lazarus:
Fixes #434.

Map `Event.transparent` to the standard `TRANSP:OPAQUE` / `TRANSP:TRANSPARENT` tokens from [RFC 5545 section 3.8.2.7](https://www.rfc-editor.org/rfc/rfc5545.html#section-3.8.2.7), using an event-specific converter. `None` still omits the property. Parsing accepts case-insensitive tokens and `VALUE` parameter names, preserves parameters, and rejects invalid values, duplicate properties and multiple type entries with differing name casing. Shared TEXT/BOOLEAN converters are unchanged. The [32 transparency regression cases](https://github.com/ics-py/ics-py/blob/6bf7fdc5251cc0fe181cf1e76509e7d4c5d6ecd3/tests/issues/gh434.py) pass.

The repeated hosted failures also needed prerequisites, kept in separate follow-up commits:

- `592db0f`: remove the failing `lipsum` dependency while retaining all five long-text fixture scales, and fix README formatting, following C4ptainCrunch's existing changes in draft #441. Select Python 3.12 for the existing pre-commit hooks and pin stubs compatible with the existing mypy 0.960. Correct the [TEXT test helper](https://github.com/ics-py/ics-py/blob/6bf7fdc5251cc0fe181cf1e76509e7d4c5d6ecd3/tests/valuetype/text.py) to unfold physical lines, add a deterministic folded-text example, make the clock comparison deterministic, and retain installed-package assertions using portable paths. The broader release changes from #441 are not included.
- `6bf7fdc`: resolve omitted timezone alias files through the existing dateutil dependency, then load the canonical bundled VTIMEZONE. Direct resources and Windows mappings retain priority; unknown IDs still fail. [Alias tests](https://github.com/ics-py/ics-py/blob/6bf7fdc5251cc0fe181cf1e76509e7d4c5d6ecd3/tests/timezone.py) cover winter/summer offsets, named and absolute Unix-style paths, lookup precedence and missing data. The existing US/Pacific round-trip test remains and now serializes `event2` for its second comparison. Its regression batch changes from 9 failures / 7 passes to 16 passes.

Validation at `6bf7fdc` on Windows / Python 3.12.14:

- Fresh dependency installation succeeds. Full pytest: **371 passed, 2 failed**, with no tests skipped: the remaining doctest failures are the existing pre-epoch local-time conversion on Windows and unavailable POSIX `time.tzset`.
- Mypy checks all 34 source files successfully with the declared stub pins. Flake8, all eight hooks on changed files, wheel/sdist build and strict Twine checks pass. Installed and packaged source matches the checkout.
- The full hook run retains the Windows checkout's materialized changelog-symlink EOF issue; the symlink itself is unchanged. Sphinx completes with 29 warnings. The new third-party `dateutil.zoneinfo` import has a scoped typing annotation because the pinned stubs omit that module.

Hosted verification at `6bf7fdc`: the [Ubuntu matrix and package build](https://github.com/ics-py/ics-py/actions/runs/35405973996) and [pre-commit](https://github.com/ics-py/ics-py/actions/runs/35405973954) now pass. All seven matrix jobs succeed (CPython 3.8–3.12 and PyPy 3.8–3.9); the Python 3.11 and 3.12 logs each record **373 passed**. Python 3.11 also completes the configured typing checks and documentation build, with four documentation warnings. PyPI publication is skipped for this pull request, as configured. Required changelog and author entries are included.


x0Lazarus:
The current [test/package workflow](https://github.com/ics-py/ics-py/actions/runs/35405973996) and [pre-commit](https://github.com/ics-py/ics-py/actions/runs/35405973954) pass at `6bf7fdc`, but three required checks have not reported: `Test with Python 3.13`, `Test with Python 3.14`, and `docs/readthedocs.org:icspy`.

The current workflow and tox configuration only define Python 3.8 through 3.12 and PyPy jobs; #441 proposes the newer runtimes as part of its broader update. Could you confirm whether 3.13/3.14 should be added here or handled through #441, and check whether Read the Docs PR reporting is enabled? The missing external status is separate from the successful documentation build in the Python 3.11 job.

C4ptainCrunch:
Hi,

Your PR seems to fix multiple things at once, could you split it in multiple PRs ?
Also, while i understand the use of an LLM to write code, [I'd like that the time I spend reviewing PRs is spent talking to humans and growing their understanding of the codebase](https://godotengine.org/article/contribution-policy-2026/). So please, write your PR descriptions by hand and disclose your use of LLMs while writing the code.

